PR #20, opened 2021-05-27
making it possible to extend UCERF3_GridSourceGenerator

voj:
These changes enable us to extend `UCERF3_GridSourceGenerator` and write our own initialisation code in the new constructor.

### DIFF
--- a/src/main/java/scratch/UCERF3/griddedSeismicity/UCERF3_GridSourceGenerator.java
+++ b/src/main/java/scratch/UCERF3/griddedSeismicity/UCERF3_GridSourceGenerator.java
@@ -34,21 +34,21 @@ import com.google.common.collect.Maps;
  */
 public class UCERF3_GridSourceGenerator extends AbstractGridSourceProvider {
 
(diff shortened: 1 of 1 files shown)


kevinmilner:
@voj is there a reason that you need to extend it? This is a to-be-thrown-away (or heavily refactored) class, might be better in the short-term to just copy and modify it for your uses for now. Otherwise you'll be more affected if this class disappears or is heavily modified.

voj:
> 
> 
> @voj is there a reason that you need to extend it? This is a to-be-thrown-away (or heavily refactored) class, might be better in the short-term to just copy and modify it for your uses for now. Otherwise you'll be more affected if this class disappears or is heavily modified.

@kevinmilner the idea was to duplicate as little code as possible. If you think copying it is safer than I'll do that. But I suppose when this is thrown out or heavily refactored we'll probably have to do the same anyway?

voj:
Closing this to follow @kevinmilner 's recommendation of duplicating the code instead

PR #21, opened 2021-05-28
Modular Fault System Rupture Set/Solution

kevinmilner:
This is a working draft of my idea for how to re-think the nasty hierarchy of FaultSystemRupSet/Solution classes. I've given it some thought and settled on this approach after chasing other dead ends, but it's still definitely up for debate and the specifics of the API are very much subject to change based on feedback. Tagging the most likely interested parties here so you can provide feedback and track the progress: @field-usgs @chrisbc @voj @pmpowers-usgs 

I'll start with the motivation:

## Current limitations

The class hierarchy for solutions and rupture sets is complex as is, and without a redesign it's about to get way worse as we begin UCERF4 development (and NZ integration). Here's some of it as is (does not include the NZ stuff):

* **FaultSystemRupSet**: base class
  * _SlipEnabledRupSet_: interface that adds average-slip functionality
    * _SlipAlongRuptureModelRupSet_: interface that adds Dsr functionality
      * **InversionFaultSystemRupSet**: UCERF3 rupture set class
* **FaultSystemSolution**: base class
  * _SlipEnabledSolution_: interface that adds solution slip rate calculation functionality
    * **InversionFaultSystemSolution**: UCERF3 solution
      * **AverageFaultSystemSolution**: Average of multiple inversions
  * **SimulatorFaultSystemSolution**: RSQSim, etc

That's a lot as it is, and Kiwis will want to add many analogous classes. There are properties of the UCERF3 classes, on top of the existing interfaces, that we will want to carry forward and have show up in plots/diagnostic pages, such as:

* Logic tree branch information
* Information about plausibility filters used to create a rupture set
* Inversion metadata, misfits
* Cluster Ruptures
* Branch-averaged information, such as rupture MFDs
* Grid source provider

In the future, lets say we wanted an 'average' UCERF4 fault system solution. We would first have to extract an 'AverageSolution' interface that lies between InversionFaultSystemSolution and AverageFualtSystemSolution, and make another sub-class of UCERF4 solutions that contain average functionality. Now we have yet another interface with 2 additional implementations.

Things get even nastier when you're trying to load in a rupture set/solution from a file. Currently, I have to figure out which of the concrete classes that I know about best matches the information available in the zip file. This gets tough, and only gets more complicated as more and more classes/interfaces are added. If you want to see the nastiness, [here's the current code for loading a rupture set](https://github.com/opensha/opensha/blob/a62c41ecca27cbb786f76b78e8c5c01ff464ad65/src/main/java/scratch/UCERF3/utils/FaultSystemIO.java#L219). Now imagine that the Kiwis want to add custom information to their rupture sets that needs to be serialized to a rupture set zip file: they would have to add a bunch more special cases to our existing I/O class in order to do that. Yikes.

## Modular Proposal

I propose that we strip RupSet and Solution to their basic capabilities, those that are common all all conceivable implementations. Any additional features are then represented as modules, that can be loaded into and retrieved from any rupture set/solution. There are different [RupSetModule](https://github.com/opensha/opensha/blob/modular-fault-sys-rup-set/src/main/java/org/opensha/sha/earthquake/faultSysSolution/modules/RupSetModule.java) and [SolutionModule](https://github.com/opensha/opensha/blob/modular-fault-sys-rup-set/src/main/java/org/opensha/sha/earthquake/faultSysSolution/modules/SolutionModule.java) base classes. Those classes really only force that modules have a name, and gives them access to the rupture set/solution to which they belong.

You then retrieve a module by asking a rupture set/solution for it using the class. For example, lets say you have a module called LogicTreeBranchModule that supplies branch information to rupture sets. First, you can register it with a rupture set:

```
rupSet.addModule(new LogicTreeBranchModule(rupSet, branch));
```

You can then retrieve it by calling:

```
LogicTreeBranchModule branchModule = rupSet.getModule(LogicTreeBranchModule.class);
```

Or query for it's existence by calling

```
rupSet.hasModule(LogicTreeBranch.class);
```

Modules can be subclassed, and if so than you can call for any super-class when retrieving a module. So if you have a `MyLogicTreeBranchModule` that extends `LogicTreeBranchModule`, you can retrieve it with either `rupSet.getModule(LogicTreeBranchModule.class);` or `rupSet.getModule(MyLogicTreeBranchModule.class);`.

## I/O Implications

One of the main benefits of this is that it can drastically simplify I/O. By default, modules are transient (not written to files), but they can become stateful by implementing the [StatefulModule](https://github.com/opensha/opensha/blob/modular-fault-sys-rup-set/src/main/java/org/opensha/sha/earthquake/faultSysSolution/modules/StatefulModule.java) interface. Modules themselves are responsible for their (de)serialization, so anyone can add their own modules an use the same I/O code (so long as their module is on the classpath). I have added helper interfaces for [JSON backed modules](https://github.com/opensha/opensha/blob/modular-fault-sys-rup-set/src/main/java/org/opensha/sha/earthquake/faultSysSolution/modules/JSON_BackedModule.java) and [CSV backed modules](https://github.com/opensha/opensha/blob/modular-fault-sys-rup-set/src/main/java/org/opensha/sha/earthquake/faultSysSolution/modules/CSV_BackedModule.java) where most of the work is done for you.

Here's a super-simple example SolutionModule implementation that is backed by a CSV. The new I/O code will automatically store the CSV to a solution zip file, and load the module from it:

```java
public class RupMFDsModule extends SolutionModule implements CSV_BackedModule {
	
	private DiscretizedFunc[] rupMFDs;
	
	public RupMFDsModule() {
		super(null);
	}

	public RupMFDsModule(FaultSystemSolution sol, DiscretizedFunc[] rupMFDs) {
		super(sol);
		this.rupMFDs = rupMFDs;
		int numRups = sol.getRupSet().getNumRuptures();
		Preconditions.checkState(numRups == rupMFDs.length,
				"Have %s ruptures but %s rupture MFDs", numRups, rupMFDs.length);
	}

	@Override
	public String getName() {
		return "Rupture MFDs";
	}

	@Override
	public String getCSV_FileName() {
		return "rup_mfds.csv";
	}

	@Override
	public CSVFile<?> getCSV() {
		CSVFile<String> csv = new CSVFile<>(true);
		csv.addLine("Rupture Index", "Magnitude", "Rate");
		// do stuff here
		return csv;
	}

	@Override
	public void initFromCSV(CSVFile<String> csv) {
		// do stuff here
	}
	
	public DiscretizedFunc getRuptureMFD(int rupIndex) {
		return rupMFDs[rupIndex];
	}

}
```

I'm also taking this opportunity to revamp the file structure, switching out the old binary formats for simpler CSVs. I'll edit this description to provide more details about that in the future.

## Current questions/TODOs

* Should the rupture set/solution classes be final? Making them final forces us to follow the design pattern, and things could get messy if not
* I'm interested in retiring the fault system XML format, but not sure what the format should be. It could be represented in CSVs, or in GeoJSON, though we would need to add a bunch of custom fields to GeoJSON for it to work.
* In order to be deserialized, all modules currently require a public no-arg constructor. There is no way to enforce this at compile-time, however, so that gets a little nasty. Any suggestions on improving this?

### DIFF
--- a/src/main/java/org/opensha/sha/earthquake/faultSysSolution/FaultSystemRupSet.java
+++ b/src/main/java/org/opensha/sha/earthquake/faultSysSolution/FaultSystemRupSet.java
@@ -1,9 +1,5 @@
-/**
- * 
- */
-package scratch.UCERF3;
+package org.opensha.sha.earthquake.faultSysSolution;
 
-import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -11,22 +7,17 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.Callable;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
+
+import javax.annotation.Nullable;
 
 import org.apache.commons.math3.stat.StatUtils;
-import org.opensha.commons.calc.FaultMomentCalc;
 import org.opensha.commons.geo.LocationList;
 import org.opensha.commons.geo.Region;
 import org.opensha.commons.geo.RegionUtils;
-import org.opensha.commons.util.ExceptionUtils;
-import org.opensha.sha.earthquake.faultSysSolution.ruptures.ClusterRupture;
-import org.opensha.sha.earthquake.faultSysSolution.ruptures.FaultSubsectionCluster;
-import org.opensha.sha.earthquake.faultSysSolution.ruptures.plausibility.PlausibilityConfiguration;
-import org.opensha.sha.earthquake.faultSysSolution.ruptures.util.RuptureConnectionSearch;
+import org.opensha.sha.earthquake.faultSysSolution.modules.ModuleManager;
+import org.opensha.sha.earthquake.faultSysSolution.modules.RupSetModule;
+import org.opensha.sha.earthquake.faultSysSolution.modules.impl.ClusterRuptures;
+import org.opensha.sha.earthquake.faultSysSolution.modules.impl.PlausibilityConfigurationModule;
 import org.opensha.sha.faultSurface.CompoundSurface;
 import org.opensha.sha.faultSurface.FaultSection;
 import org.opensha.sha.faultSurface.FaultTrace;
@@ -40,40 +31,39 @@ import com.google.common.collect.Maps;
 import com.google.common.collect.Table;
 import com.google.common.collect.Table.Cell;
 
-import scratch.UCERF3.analysis.DeformationModelsCalc;
-
-
 /**
  * This class represents the attributes of ruptures in a fault system, 
  * where the latter is composed of some number of fault sections.
+ * <p>
+ * Only the core fields common to rupture sets from all models are included. Extra attributes can be attached to this
+ * rupture set as RupSetModule instances. Examples of modules include: logic tree branches, gridded seismicity,
+ * plausibility configurations, cluster rupture representations, RSQSim event mappings
  * 
  * @author Field, Milner, Page, & Powers
  *
  */
-@Deprecated
-public class FaultSystemRupSet implements Serializable {
+public final class FaultSystemRupSet {
+// TODO: should this be final?
 	
 	// data arrays/lists
 	private List<? extends FaultSection> faultSectionData;
 	private double[] mags;
 	private double[] sectSlipRates;
-	private double[] sectSlipRateStdDevs;
+	private double[] sectSlipRateStdDevs; // TODO: make a module?
 	private double[] rakes;
 	private double[] rupAreas;
 	private double[] rupLengths;
 	private double[] sectAreas;
 	private List<List<Integer>> sectionForRups;
 	private String info;
-
-	private PlausibilityConfiguration plausibilityConfig;
-	private List<ClusterRupture> clusterRuptures;
 	
-	// for caching
+	// if true, caching operations will show a graphical progress bar
 	protected boolean showProgress = false;
 	
-	// NOTE: copy param documentation to init() method if you make any changes below
+	private ModuleManager<RupSetModule> moduleManager;
+	
 	/**
-	 * Constructor for precomputed data where everything is passed in.
+	 * Initialized a FaultSystemRupSet object with all core data.
 	 * 
 	 * @param faultSectionData fault section data list (CANNOT be null)
 	 * @param sectSlipRates slip rates for each fault section with any reductions applied (CAN be null)
@@ -88,65 +78,15 @@ public class FaultSystemRupSet implements Serializable {
 	 */
 	public FaultSystemRupSet(
 			List<? extends FaultSection> faultSectionData,
-			double[] sectSlipRates,
-			double[] sectSlipRateStdDevs,
-			double[] sectAreas,
+			@Nullable double[] sectSlipRates,
+			@Nullable double[] sectSlipRateStdDevs,
+			@Nullable double[] sectAreas,
 			List<List<Integer>> sectionForRups,
 			double[] mags,
 			double[] rakes,
 			double[] rupAreas,
-			double[] rupLengths,
-			String info) {
-		init(faultSectionData, sectSlipRates, sectSlipRateStdDevs, sectAreas,
-				sectionForRups, mags, rakes, rupAreas, rupLengths, info);
-	}
-	
-	/**
-	 * Default constructor for subclasses which will call init on their own.
-	 * 
-	 * Protected so it can only be invoked by subclasses.
-	 */
-	protected FaultSystemRupSet() {
-		// do nothing, it's up to subclass to call init.
-	}
-	
-	/**
-	 * Initialize from another rupSet
-	 * @param rupSet
-	 */
-	protected void init(FaultSystemRupSet rupSet) {
-		init(rupSet.getFaultSectionDataList(), rupSet.getSlipRateForAllSections(),
-				rupSet.getSlipRateStdDevForAllSections(), rupSet.getAreaForAllSections(),
-				rupSet.getSectionIndicesForAllRups(), rupSet.getMagForAllRups(), rupSet.getAveRakeForAllRups(),
-				rupSet.getAreaForAllRups(), rupSet.getLengthForAllRups(), rupSet.getInfoString());
-		copyCacheFrom(rupSet);
-	}
-	
-	/**
-	 * Sets all parameters
-	 * 
-	 * @param faultSectionData fault section data list (CANNOT be null)
-	 * @param sectSlipRates slip rates for each fault section with any reductions applied (CAN be null)
-	 * @param sectSlipRateStdDevs slip rate std deviations for each fault section (CAN be null)
-	 * @param sectAreas areas for each fault section (CAN be null)
-	 * @param sectionForRups list of fault section indexes for each rupture (CANNOT be null)
-	 * @param mags magnitudes for each rupture (CANNOT be null)
-	 * @param rakes rakes for each rupture (CANNOT be null)
-	 * @param rupAreas areas for each rupture (CANNOT be null)
-	 * @param rupLengths lengths for each rupture (CAN be null)
-	 * @param info metadata string
-	 */
-	protected void init(
-			List<? extends FaultSection> faultSectionData,
-			double[] sectSlipRates,
-			double[] sectSlipRateStdDevs,
-			double[] sectAreas,
-			List<List<Integer>> sectionForRups,
-			double[] mags,
-			double[] rakes,
-			double[] rupAreas,
-			double[] rupLengths,
-			String info) {
+			@Nullable double[] rupLengths,
+			@Nullable String info) {
 		Preconditions.checkNotNull(faultSectionData, "Fault Section Data cannot be null");
 		this.faultSectionData = faultSectionData;
 		Preconditions.checkNotNull(faultSectionData, "Magnitudes cannot be null");
@@ -182,6 +122,8 @@ public class FaultSystemRupSet implements Serializable {
 		this.sectionForRups = sectionForRups;
 		
 		this.info = info;
+		
+		this.moduleManager = new ModuleManager<>(RupSetModule.class);
 	}
 	
 	/**
@@ -246,28 +188,6 @@ public class FaultSystemRupSet implements Serializable {
 	}
 	
 	/**
-	 * This returns the magnitude of the smallest rupture involving this section or NaN
-	 * if no ruptures involve this section.  This is called "Orig" because subclasses
-	 * may filter the minimum magnitudes further (e.g., so they don't fall below some
-	 * threshold).
-	 * @param sectIndex
-	 * @return
-	 */
-	public double getOrigMinMagForSection(int sectIndex) {
-		List<Integer> rups = getRupturesForSection(sectIndex);
-		if (rups.isEmpty())
-			return Double.NaN;
-		double minMag = Double.POSITIVE_INFINITY;
-		for (int rupIndex : getRupturesForSection(sectIndex)) {
-			double mag = getMagForRup(rupIndex);
-			if (mag < minMag)
-				minMag = mag;
-		}
-		return minMag;
-	}
-	
-	
-	/**
 	 * This returns the magnitude of the largest rupture involving this section or NaN
 	 * if no ruptures involve this section.
 	 * @param sectIndex
@@ -284,104 +204,6 @@ public class FaultSystemRupSet implements Serializable {
 				maxMag = mag;
 		}
 		return maxMag;
-	}
-	
-	
-	/**
-	 * This computes the fractional slip rate (or moment rate) taken away for sub-seismogenic ruptures
-	 * (relative to creep reduced moment rate).  Actually, this will also include any additional coupling
-	 * coefficient reductions applied by subclasses (e.g., InversionFaultSystemRuptureSet has the option
-	 * of also applying an implied coupling coefficient that will be reflected in getSlipRateForSection(sectIndex))
-	 * @param sectIndex
-	 * @return
-	 */
-	public double getMomentRateReductionFraction(int sectIndex) {
-		double origSlipRate = getFaultSectionData(sectIndex).getReducedAveSlipRate() * 1e-3; // convert to meters
-		double reducedSlipRate = getSlipRateForSection(sectIndex);
-		return 1d - reducedSlipRate/origSlipRate;
-	}
-	
-	/**
-	 * This returns the total reduction in moment rate for subseimogenic ruptures
-	 * and any coupling coefficient applied (the amount removed).  Actually, this 
-	 * reduction also includes any additional coupling coefficient reductions applied 
-	 * by subclasses (e.g., InversionFaultSystemRuptureSet has the option of also 
-	 * applying an implied coupling coefficient that will be reflected in 
-	 * getSlipRateForSection(sectIndex))
-	 * 
-	 * @return
-	 */
-	public double getTotalMomentRateReduction() {
-		return getTotalOrigMomentRate() - getTotalReducedMomentRate();
-	}
-	
-	/**
-	 * This returns the total fraction of moment that is reduced by subseismogenic ruptures.
-	 * Actually, this reduction also includes any additional coupling coefficient reductions  
-	 * applied by subclasses (e.g., InversionFaultSystemRuptureSet has the option of also 
-	 * applying an implied coupling coefficient that will be reflected in 
-	 * getSlipRateForSection(sectIndex))
-	 * 
-	 */
-	public double getTotalMomentRateReductionFraction() {
-		return getTotalMomentRateReduction() / getTotalOrigMomentRate();
-	}
-
-	/**
-	 * This returns the original moment rate (with creep reductions but without subseismogenic
-	 * rupture reductions) for a fault subsection
-	 */
-	public double getOrigMomentRate(int sectIndex) {
-		FaultSection sectData = getFaultSectionData(sectIndex);
-		double moRate = sectData.calcMomentRate(true);
-		if (Double.isNaN(moRate))
-			return 0;
-		return moRate;
-	}
-	
-	/**
-	 * This returns the total moment rate for the given rupSet without taking into account any
-	 * moment rate reductions for subseismogenic ruptures (but does include all default creep reductions).<br>
-	 * <br>
-	 * This simply calls <code>DeformationModelsCalc.calculateTotalMomentRate(sectData, true)</code> 
-	 * 
-	 * @param rupSet
-	 * @return
-	 */
-	public double getTotalOrigMomentRate() {
-		return DeformationModelsCalc.calculateTotalMomentRate(getFaultSectionDataList(), true);
-	}
-	
-	/**
-	 * This returns the moment rate after removing that for subseimogenic ruptures 
-	 * (and default creep effects). This also include any additional
-	 * coupling coefficients applied by subclasses (e.g., InversionFaultSystemRuptureSet 
-	 * has the option of also applying an implied coupling coefficient that will be 
-	 * reflected in getSlipRateForSection(sectIndex)).
-	 * 
-	 * @param sectIndex
-	 * @return
-	 */
-	public double getReducedMomentRate(int sectIndex) {
-		return getOrigMomentRate(sectIndex) * (1 - getMomentRateReductionFraction(sectIndex));
-	}
-	
-	/**
-	 * This returns the total moment rate after removing that for subseismogenic  
-	 * ruptures (and default creep influences).  This also include any additional
-	 * coupling coefficients applied by subclasses; e.g., InversionFaultSystemRuptureSet 
-	 * has the option of also applying an implied coupling coefficient that will be 
-	 * reflected in getSlipRateForSection(sectIndex)).
-	 * @return
-	 */
-	public double getTotalReducedMomentRate() {
-		double totMoRate = 0d;
-		for (int sectIndex=0; sectIndex<getNumSections(); sectIndex++) {
-			double sectMoment = getReducedMomentRate(sectIndex);
-			if (!Double.isNaN(sectMoment))
-				totMoRate += sectMoment;
-		}
-		return totMoRate;
 	}
 	
 	/**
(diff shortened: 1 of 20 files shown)


voj:
Hi @kevinmilner, here are my thoughts:

I'm not sure I fully understand the motivation. That class and interface hierarchy does not look overly complicated to me for a Java library. Are you worried you might end up with a proliferation of classes that would have been simplified by multiple inheritance or mixins?

I agree that loading from file is complicated at the moment, but I don't think it needs to be. There should not be a monolithic, hardcoded, method that loads all types of files. When we write a file, we should know what it should be loaded as. So we should simply write that information into the file. We could have a generic loader that our classes can register with, and each class knows how to load itself from file. This is how most parsers work, for example https://sites.google.com/site/gson/gson-user-guide#TOC-Custom-Serialization-and-Deserialization

Having said that, I do like some of your suggestions. Let's have a chat tomorrow :-)

chrisbc:
@kevinmilner @voj thankyou for the time and thought put into this. I've seen some rather large and intractible codebases  in commercial systems (e.g. >1000 'production' repos a Xro). So,  by comparison I think we have a reasonably manageable set of problems to deal with here :). Perhaps most important is to understand the issues and their impact on science productivity.

kevinmilner:
@voj, thanks a lot for the comments! Lets definitely chat more on our call later today, but I’ll reply here publicly for other’s benefit.

I agree that the current class hierarchy isn’t insane, but it will quickly balloon as we try to share information between and write common code that can use U3/U4/NZ rupture sets/solutions. Here’s what it might look like in the future on the present course:

Rupture Set Interfaces:

* _SlipEnabledRupSet_: adds average-slip functionality
  * _SlipAlongRuptureModelRupSet_: adds Dsr functionality, requires average slip
* _LogicTreeBranchProvider_: adds logic tree branch info
* _InversionConstraintProvider_: adds inversion constraint info
* _ClusterRuptureSet_: adds 'cluster rupture' representations, required for diagnostic pages
* _PlausibilityConfigProvider_: adds information about the plausibility filters used to create this rupture set
* ...surely more as time goes on

Solution interfaces:

* _GridSourceProvider_: adds gridded seismicity information
* _InversionMisfits_: adds info about inversion misfits, metadata
* _AveFaultSystemSolution_: adds ability to get the rate of the rth rupture in the nth inversion run
* ...surely more as time goes on

So then our pipeline goes something like this:

1. ClusterRupture builds a rupture set that implements _ClusterRuptureSet_ and _PlausibilityConfigProvider_. We probably want to be able to save that to a file, so we need a logic to save/load this type of file
2. We then want to configure an inversion using those ruptures, so we have a factory class of some sort that adds _LogicTreeBranchProvider_, _Slip*RupSet_, and _InversionConstraintProvider_. This is where it stats to gets messy: we need to create a new wrapper class that implements the new interfaces, and also still implements the super interfaces. We'll want to be able to save/load this type of pre-inversion file as well.
3. Run an inversion, return a FaultSystemSolution that implements _InversionMisfits_, probably write that out to a file as part of a batch job on a cluster
4. Load that inversion in, add _GridSourceProvider_ information
5. Optionally also load in multiple inversion runs, create an _AveFaultSystemSolution_ class and support I/O

The main issues that I see with this pipeline are:

1. Every time the next step in the pipeline adds additional information, it needs to create a new concrete class that implements its interface and also allows instanceof commands to work on all of the interfaces from the passed in rupture set/solution, e.g., step 2 above. This works if we're always passing in the same inputs, but we might want to run an inversion on a rupture set that was built externally (e.g., RSQSim ruptures) that doesn't implement the same interfaces, further complicating the code at this stage.
2. I/O becomes a nightmare, because for each unique set of inputs we need to know what concrete base class to use. This can be mitigated, as you suggest, by defining some sort of loader interface and making each rup set/solution define the class that can be used to load it. That breaks down if I'm trying to load a rupture set built with a downstream project and don't have your code on my classpath, however, such as loading it in SCEC-VDO or the OpenSHA hazard calculation GUI apps that won't depend on the NZ codebase. In that case we could just default to the basic rupture set/solution loading code and keep only core functionality, but it creates a ton of new loader/writer classes at every step along the way.

All of this would be much simpler if java had a way to truly inject an interface at runtime, e.g., by saying I have class A that implements B, and I want to now at runtime also let it implement C (and still implement B). Without that, we need a unique concrete class and I/O helper class for every unique combination of interfaces that we'll ever encounter. I first tried to go down this route with some really dirty java using the Proxy class and a custom ClassLoader to dynamically inject interfaces at Runtime, and got it to work but everything takes a performance hit there (all method calls require reflection) and it was just nasty.

Instead, the proposed modular system allows each step in the pipeline to add information without creating a new concrete subclass, and allows for a single point of loading that lets each module load and save its own data (or skip with a warning if the loading class isn't found on the current classpath). We still have lots of extra little concrete classes (the modules), but don't have to add extra classes for each combination therein and fancy loading logic to always load in a class that implements the correct set of interfaces.

The main downside that I see of the modular system is that it doesn't use the `instanceof` keyword, so instead of the simpler and more standard `rupSet instanceof SlipEnabledRupSet` we would do `rupSet.hasModule(AveSlipData.class)`, which is less intuitive. I'm certainly open to suggestions, lets discuss this afternoon.

chrisbc:
@kevinmilner thanks for all the extra detail . Just a couple of thoughts/questions....

> The main issues that I see with this pipeline are:

>    1.    Every time the next step in the pipeline adds additional information, it needs to create a new concrete class that implements its interface and also allows instanceof commands to work on all of the interfaces from the passed in rupture set/solution, e.g., step 
>   2 above.**This works if we're always passing in the same inputs**, but we might want to run an inversion on a rupture set that was built externally (e.g., RSQSim ruptures) that doesn't implement the same interfaces, further complicating the code at this stage.

So perhaps these edge cases could just adapt to the 'standard' form using some adapter code [in their codebase :)]. IE it is the responsibility of the external actor/project to conform to the opensha Class model. 

>    2.    I/O becomes a nightmare, because for each unique set of inputs we need to know what concrete base class to use. This can  be mitigated, as you suggest, by defining some sort of loader interface and making each rup set/solution define the class that   can be used to load it. That breaks down if I'm trying to load a rupture set built with a downstream project and **don't have your code on my classpath**, however, such as loading it in SCEC-VDO or the OpenSHA hazard calculation GUI apps that won't depend on the NZ codebase. In that case we could just default to the basic rupture set/solution loading code and keep only core functionality, but it creates a ton of new loader/writer classes at every step along the way.

If we follow the encapsulated I/O approach, then the runtime problem becomes 'get the code on the classpath'. But that's a tractable one?

chrisbc:
And, while we're on the subject of **I/O,** whatever we come up with I suggest we give more weight to schema support - which is possible with both XML and JSON formats. I think this could alleviate a good deal of integration pain, making it faster/safer for science teams to harness & combine their software tools.

NB The current FaultModel XML is not schema-compatible (it has dynamic element name) - so that has to change.

voj:
@kevinmilner as promised here are some suggestions:

I think it's best to separate concerns as much as possible. I would define a rupture set as a module, not as the foundation for the file. I would also completely avoid having any registry that takes care of loading modules and working out which module in the class hierarchy you might want when you call `get` on the file.

In pseudo-Java:

```
public interface OpenShaFileModule{
    public void save(ZipFile file);
}

public class OpenShaFile{
    protected ZipFile file;
    protected Map<String,OpenShaFileModule> moduleCache;
    
    public OpenShaFile(String fileName){
        // load zipfile
        moduleCache = new HashMap<>(); // modules are loaded lazily
    }
    
    public void save(String fileName){
        //1. copy contents of existing zipfile
        //2. save each module
    }
    
    public ZipFile getFile(){return file};
    
    public OpenShaFileModule getModuleFromCache(String name){
        return moduleCache.get(name);
    }
    
    public void addModule(String name, OpenShaFileModule module){
        moduleCache.put(name, module);
    }
}

public class RuptureSetModule implements OpenShaFileModule{
    
    static final String name = "RuptureSet";
    
    public static RuptureSetModule getFrom(OpenShaFile file){
        RuptureSetModule module = (RuptureSetModule) file.getModuleFromCache(name);
        if(module != null){
            return module;
        }else{
            // do all the file loading and object initialisation
            // put module into cache
            // return module
        }
    }

    @Override
    public void save(ZipFile file) {
        // write data to zipfile
    }
    
    // actual rupture set methods
}

```

This way, everything is kept as simple as possible. To get a RuptureSetModule, you call `RuptureSetModule.getFrom(openshaFile)`

When you have constructed a rupture set from a fault model, you could call something like `openshaFile.addModule(ruptureSet)`

Note that this is not symmetrical, but would make the code much simpler. There is no central code that needs to know or manage what is in the file and how it is loaded and stored. 

My other suggestion was to implement this and then to use it as the back-end to the existing RuptureSet and FaultSystemSolution implementations. This means legacy code like UCERF3 does not need to be updated, the new file format would work right away, and you're sure that everything works as expected. It then gives you all the time in the world to update the sections of the code base that you care about.

kevinmilner:
@voj (and cc @chrisbc) thanks much for your suggestions, they were really helpful. I like your approach a lot and ran with it, but probably added more complexity than you were after in my implementation. I think it's better than what I had before though, and is quite flexible.

So far, I have just posted a demo implementation [here in opensha-dev](https://github.com/opensha/opensha-dev/tree/master/src/main/java/scratch/kevin/ucerf3/inheritenceSandbox/modular2). Here are the basic classes:

## Modules themselves (& helpers)

* _OpenSHA_Module_: basically just a tagging interface for a module, does nothing other than extend Named
  * _ArchivableModule_: adds methods for storing/loading module information to/from a zip archive
    * _FileBackedModule_: provides default implementations for the above, requiring that implementations just supply a file name and ability to write to/init from streams
      * _CSV_BackedModule_: same, but now to/from a CSV
      * _JSON_BackedModule_: same, but now to/from a JSON
      * _TextBackedModule_: same, but now to/from plain text

## Module containers

Modules can be stored in and retrieved from containers. The base class for this is **ModuleContainer** which provides `addModule(OpenSHA_Module)`, `hasModule(Class<? extends OpenSHA_Module)`, and `getModule(Class<? extends OpenSHA_Module)`. I'm back to using the classes here because I find Strings to be messy. I think we still need a registry, as a rupture set can contain many sub modules, but some modules aren't applicable to some rupture sets, so we need a way to ask a rupture set if it has a particular module or not.

Module inheritance is handled as before. Lets say you have these module classes:

```
	abstract class AbstractCustomModule implements OpenSHA_Module {
		// abstract module code
	}
	
	class CustomModuleImpl extends AbstractCustomModule {
		// concrete module code
	}
```

If you add a module that is of type `CustomModuleImpl`, you can retrieve it via either calling `getModule(CustomModuleImpl.class)` or `getModule(AbstractCustomModule.class)`, as both the base class and concrete implementation are of type OpenSHA_Module and thus would be mapped to the registered instance. Helper interfaces that should never be registered to an can be tagged with the `@ModuleHelper`. This prevents things like the _CSV_BackedModule_ helper interface from ever being mapped to an implementation.

**ModuleArchive** extends **ModuleContainer** by adding methods to write modules to and load them from a zip archive. Modules are loaded lazily by using `container.addAvailableModule(Callable<E> call, Class<E> clazz)`.

Something can be both a module and a container, in that case sub-modules are nested in the zip file (see below).

## Demo Rupture Set and Solution implementations

Here is my demo Rupture Set class. It is both a module, and a container for other modules. Sub-modules of the rupture set will be written to the 'ruptures' sub-directory of the zip file. It can be written to and loaded from a zip file:

```
public class DemoFSRS extends ModuleContainer<OpenSHA_Module> implements ArchivableModule {

	public static DemoFSRS load(File file) throws IOException {
		ModuleArchive<OpenSHA_Module> archive = new ModuleArchive<>(file);
		
		DemoFSRS rupSet = archive.getModule(DemoFSRS.class);
		Preconditions.checkState(rupSet != null);
		rupSet.setArchive(archive);
		
		return rupSet;
	}

	public void writeArchive(File file) throws IOException {
		if (archive == null)
			archive = new ModuleArchive<>();
		if (!archive.hasModule(DemoFSRS.class))
			archive.addModule(this);
		archive.writeArchive(file);
	}

	// ...
}
```

The solution class is set up similarly:

```
public class DemoFSS extends ModuleContainer<OpenSHA_Module> implements ArchivableModule {
	public static DemoFSS load(File file) throws IOException {
		ModuleArchive<OpenSHA_Module> archive = new ModuleArchive<>(file);
		
		DemoFSRS rupSet = archive.getModule(DemoFSRS.class);
		Preconditions.checkState(rupSet != null);
		DemoFSS sol = archive.getModule(DemoFSS.class);
		Preconditions.checkState(sol != null);
		sol.setArchive(archive);
		sol.rupSet = rupSet;
		
		return sol;
	}
	
	public void writeArchive(File file) throws IOException {
		System.out.println("Writing FSS to "+file.getAbsolutePath());
		boolean existing = archive != null;
		if (archive == null)
			archive = new ModuleArchive<>();
		if (!archive.hasModule(DemoFSS.class)) {
			archive.addModule(this);
			archive.addModule(rupSet);
		}
		archive.writeArchive(file, existing);
	}
}
```

Here's what it looks like as a zip archive:

* _modules.json_: master module index, lists rupture set and solution modules
* _ruptures_: sub-directory for rupture information & sub-modules
  * _ruptures/modules.json_: rupture set sub-modules
  * _ruptures/ruptures.csv_: data file used by the rupture set module
  * _ruptures/sections.csv_: data file used by the rupture set module
  * _ruptures/info.txt_: data file used by the Info module that was added to the rupture set module
* _solution_: sub-directory for solution information & sub-modules
  * _solution/modules.json_: solution sub-modules
  * _solution/ruptures.csv_: data file used by the solution module
  * _solution/sections.csv_: data file used by the solution module
  * _solution/info.txt_: data file used by the Info module that was added to the solution module

...and here are the contents of those various module.json files:

modules.json:

```
[
  {
    "name": "Rupture Set",
    "className": "scratch.kevin.ucerf3.inheritenceSandbox.modular2.DemoFSRS",
    "path": "ruptures/"
  },
  {
    "name": "Solution",
    "className": "scratch.kevin.ucerf3.inheritenceSandbox.modular2.DemoFSS",
    "path": "solution/"
  }
]
```

ruptures/modules.json:

```
[
  {
    "name": "Info",
    "className": "scratch.kevin.ucerf3.inheritenceSandbox.modular2.DemoFSRS$DemoInfoModule",
    "path": "ruptures/"
  }
]
```

solution/modules.json:

```
[
  {
    "name": "Info",
    "className": "scratch.kevin.ucerf3.inheritenceSandbox.modular2.DemoFSRS$DemoInfoModule",
    "path": "solution/"
  }
]
```

This is nice because it keeps things separate, and allows for both side-by-side modules and recursive data structures. And all loading is now handled at the class level, so there's no need for a "FaultSystemIO" class anymore.

I know that I added a lot more complexity and functionality to your basic design, but I think this one is in pretty good shape now. Without objection, I'll move this demo implementation over to the modulear-fault-sys-rup-set branch tomorrow and start refactoring to use it. I think we'll be able to refactor the old classes to extend the new framework for now, and slowly ween ourselves off of using them. I'm available anytime today if you want to chat about any of this before your weekend starts (and I get cracking tomorrow).

voj:
@kevinmilner you're so fast! It's coming along nicely :-)

Here are my thoughts (or maybe test cases?):

- If you want to be able to support inheritance, the whole thing becomes more complex. For example:
    - What happens if I save a ModuleArchive with a NZRupSet module? Will an unmodified ScecVdo be able to load the rupset? (this was part of the original motivation)
    - Would it be possible to accidentally load a superclass and a subclass version of a module in different parts of the code? We'd have two in-memory representations of the data, leading to fun bugs.
- I feel a bit uneasy about the `writeArchive` method being on modules. I think each module should only have to worry about how to write itself, and users should not have a choice which save method to use. For example, in your example, if I create a new rupset and a new solution but then call `writeArchive` on the rupset rather than the solution, do I potentially lose the solution?

I'm trying to think how inheritance can be handled safely and predictably. Here are some rules that might help. For the examples in this list I'm using the hierarchy RupSet <- CoolRupSet <- NZCoolRupSet

- There can only ever be one instance of a class in the archive. E.g. only one RupSet. This limitation is introduced by your way of retrieving modules, but I think it makes other things easier too. I think we should also ensure that there is only ever one RupSet including descendants, i.e. if someone has added a CoolRupSet to the archive they cannot also add an NZCoolRupSet.
- If you inherit from a module, the first line in `saveModule()` (or whatever it's called) should be `super.saveModule()` (and the same goes for a read operation). This way, there will be a clear distinction in the file between the data from the super class and the subclass. This also means if an application only knows about the super class, it could still load that data. E.g. if we'd write NZCoolRupSet, then there would be three directories in the archive, "rupset", "coolrupset", and "nzcoolrupset".
- If you ask the archive for a module, the archive should return the youngest descendant. e.g. if your code knows RupSet, CoolRupSet, and NZCoolRupSet, and the archive has NZCoolRupSet (and due to the previous rule also the other two), it should return NZCoolRupSet (which includes all the data). If your code only knows RupSet and CoolRupset, and you ask for RupSet, you should get a CoolRupSet, ignoring the data for NZCoolRupSet.

And lastly, unrelated to inheritance: Make sure that data files that can grow large (e.g. ruptures) are written in a streamable format. For example, writing the data as one big JSON file could mean that a giant object needs to be constructed when reading it. A row-based format like CSV would be much safer in that case because you don;t need to parse the data all at once. I've had data explode during file read and write in other projects before. I think you're already doing that, so it's just a heads up to be consciously future-proof. 

I'm going to take a break from saying "rupSet" now :-D

kevinmilner:
Great points as always, @voj!

> What happens if I save a ModuleArchive with a NZRupSet module? Will an unmodified ScecVdo be able to load the rupset? (this was part of the original motivation)

You bring up some thorny edge cases. I was still expecting/hoping that rupture set and solution wouldn't be extended, or that at least extensions wouldn't add any extra information (and only exist as helpers to facilitate loading modules, and would eventually be replaced by builders).  So right now, within this framework, you would want a rupture set extension to override the getLoadingClass() method to point to the standard rupture set class, such that when saved to/loaded from a file it would come back as a standard rupture set (with any attached modules still available). So NZRupSet could exist and configure things, but it would only be loaded back in as a RupSet. We could enforce this in Rupture Set/Solution by overriding the declaring the getLoadingClass() and serialization methods final, thus not allowing implementations. 

> If you ask the archive for a module, the archive should return the youngest descendant. e.g. if your code knows RupSet, CoolRupSet, and NZCoolRupSet, and the archive has NZCoolRupSet (and due to the previous rule also the other two), it should return NZCoolRupSet (which includes all the data). If your code only knows RupSet and CoolRupset, and you ask for RupSet, you should get a CoolRupSet, ignoring the data for NZCoolRupSet.

Another, more complicated version would be to allow module to define a list of possible loading classes, in order or priority. This gets a little nasty, though it could be largely done via a default implementation (so classes themselves wouldn't have to always enumerate the possible loading tree). So in the case of CoolRupSet extends RupSet, NZCoolRupSet extends Cool RupSet, NZCoolRupSet would say: please try to load me as an NZCoolRupSet first, but if you don't know about NZCoolRupSet I can be loaded as a CoolRupSet, or even as a RupSet. The JSON file would have to list the full tree of possible loading candidates, because if you list NZCoolRupSet only and SCEC-VDO doesn't know that exists, it also won't know that it can be loaded as a RupSet. I think I prefer the first version where we just make getLoadingClass() final on rupture set/solution, but am curious to hear your thoughts.

> Would it be possible to accidentally load a superclass and a subclass version of a module in different parts of the code? We'd have two in-memory representations of the data, leading to fun bugs.

> I think we should also ensure that there is only ever one RupSet including descendants, i.e. if someone has added a CoolRupSet to the archive they cannot also add an NZCoolRupSet.

You also bring up a really good point about the possibility, in the current implementation, to load in multiple versions of the same module. Right now, if there is a NZRupSet module and and regular RupSet module, and you loaded in NZRupSet module first, then loaded in RupSet, both would be added. get(RupSet.class) would return the regular RupSet instance, but get(NZRupSet.class) would return the NZRupSet. I propose changing it such that if any mapping to a module is replaced by a new version, all mappings to that old version are removed. So in this case, that means that mapping a regular RupSet would replace the RupSet.class mapping, and also clear the NZRupSet.class mapping.

> I feel a bit uneasy about the writeArchive method being on modules. I think each module should only have to worry about how to write itself, and users should not have a choice which save method to use. For example, in your example, if I create a new rupset and a new solution but then call writeArchive on the rupset rather than the solution, do I potentially lose the solution?

That `writeArchive(File)` method was added for convenience to rupture set and isn't implementing a superinterface. The alternative would be to bring back a static I/O class to handle loading and such. In the current implementation, if a user loads a rupture set via the `load(File)` method and later calls the `writeArchive(File)` method, any additional modules (including an attached solution) would still be retained.

> Make sure that data files that can grow large (e.g. ruptures) are written in a streamable format. ...

Point taken. Actually, as currently implemented, I think the JSON is written in a stream and not all in memory at once, but CSV is first stored in memory. 

I'll get back to all of this tomorrow (my Monday), but wanted to get my thoughts out ahead of your Monday. Thanks again for all of the really thoughtful comments, it's super helpful to have other perspectives.

voj:
> So NZRupSet could exist and configure things, but it would only be loaded back in as a RupSet. We could enforce this in Rupture Set/Solution by overriding the declaring the getLoadingClass() and serialization methods final, thus not allowing implementations.

That could work. But do you want to treat RuptureSets and Solutions differently from other modules? That might just be pushing the problem out to the other modules. We could have the same rules for all of them and see what happens. As long as the file format supports possible future changes you should be fine (famous last words).

> I think I prefer the first version where we just make getLoadingClass() final on rupture set/solution, but am curious to hear your thoughts.

I think I'm in favour of keeping everything as simple as possible, so making classes or methods final is fine by me! However, you might want to consider writing out the class hierarchy in the meta data anyway to be future proof.

> I propose changing it such that if any mapping to a module is replaced by a new version, all mappings to that old version are removed. So in this case, that means that mapping a regular RupSet would replace the RupSet.class mapping, and also clear the NZRupSet.class mapping.

But you might still have a reference to the other mapping somewhere and use that, leading to data inconsistency. For now, maybe it's good enough to throw an exception if someone tries to reload a module with a different class. It would flush out any problems and doesn't require potentially complex code until you know it's a problem. It might be that this happens a lot, but it also might be that it never happens.

> That `writeArchive(File)` method was added for convenience to rupture set and isn't implementing a superinterface. The alternative would be to bring back a static I/O class to handle loading and such. 

You could also do something like `rupSet.getArchive().save()`, then there is no static code but still only a single point of truth.

> Thanks again for all of the really thoughtful comments, it's super helpful to have other perspectives.

Glad to help! I think this is a deceptively simple change that potentially brings in a raft of problems (mostly in connection with inheritance, I think), so it's good to talk this through.

kevinmilner:
@voj thanks for the suggestions, lets discuss more tomorrow. I have moved the code over to a new branch/pull request with some of the changes that we've discussed: #24

PR #24, opened 2021-06-08
Modular Fault System Rupture Set/Solution (v2)

kevinmilner:
This is version 2 of a working draft of my idea for how to re-think the nasty hierarchy of FaultSystemRupSet/Solution classes. There was lots of good discussion (especially with respect to the motivation) in #21, which I will close shortly.

cc: @field-usgs @chrisbc @voj @pmpowers-usgs 

I have created a new system of loading, tracking, and storing modules. This will likely be useful outside of rupture sets & solutions, so it's located in the `org.opensha.commons.util.modules` package. Here's the structure:

## Module Interfaces

In their simplest form, a module is anything that implements the tagging interface _OpenSHA_Module_. There are also a lot of helper interfaces, here's the structure:

* _OpenSHA_Module_: basically just a tagging interface for a module, does nothing other than extend Named
  * _ArchivableModule_: adds methods for storing/loading module information to/from a zip archive
    * _FileBackedModule_: provides default implementations for the above, requiring that implementations just supply a file name and ability to write to/init from streams
      * _CSV_BackedModule_: same, but now to/from a CSV
      * _JSON_BackedModule_: same, but now to/from a JSON
      * _TextBackedModule_: same, but now to/from plain text
  * _SubModule_<E extends _ModuleContainer_<?>>: a module that knows about its container, and through generics can enforce that its container only be a certain type

## Module Containers

Modules can be stored in and retrieved from containers. The base class for this is **ModuleContainer** which provides `addModule(OpenSHA_Module)`, `hasModule(Class<? extends OpenSHA_Module)`, and `getModule(Class<? extends OpenSHA_Module)`.

There is a subclass that handles loading module collections to/from zip files: **ModuleArchive** extends **ModuleContainer**. Here, modules are loaded lazily by using `container.addAvailableModule(Callable<E> call, Class<E> clazz)`.

Something can be both a module and a container, in that case modules are nested in the zip file (see below). Further, if a module implements the _SubModule_<E extends _ModuleContainer_<?>> interface, it will be able to access an instance of its container, including after it is loaded from a file.

### Inheritance Considerations

The container handles inheritance, and so if you have multiple possible implementations of any module and register a subclass, it will be retrievable using it's subclass or using the common base class. Further, if you can only have one instance loaded at a time for any module tree (to prevent accidental loading of multiple implementations of the same module). For example, consider these classes:

```
	abstract class AbstractCustomModule implements OpenSHA_Module {
		// abstract module code
	}
	
	class CustomModuleImpl extends AbstractCustomModule {
		// concrete module code
	}
	
	class AltModuleImpl extends AbstractCustomModule {
		// alt module code
	}
```

If you add a module that is of type `CustomModuleImpl`, you can retrieve it via either calling `getModule(CustomModuleImpl.class)` or `getModule(AbstractCustomModule.class)`, as both the base class and concrete implementation are of type OpenSHA_Module and thus would be mapped to the registered instance. If you later add a module of type `AltModuleImpl`, it will first evict the `CustomModuleImpl` (as they both share the same parent module class, `AbstractCustomModule`).

Helper interfaces that should never be registered to a module can be tagged with the `@ModuleHelper`. This prevents things like the _CSV_BackedModule_ helper interface from ever being mapped to an implementation, and prevents unrelated modules from being evicted if you add two modules that use the same helper.

## Rupture Set & Solution Implementations

Rupture Sets and Solutions are both modules and containers for sub-modules. The basic structure is as follows, but note that there is no inheritance here:

* ModularArchive: archive that contains top level modules (rupture set and solution
  * FaultSystemRupSet: rupture set module
    * sub modules
  * FaultSystemSolution: solution module
    * sub modules

Here are the class definitions:

```
package org.opensha.sha.earthquake.faultSysSolution;

public class FaultSystemRupSet extends ModuleContainer<OpenSHA_Module> implements ArchivableModule,
  SubModule<ModuleArchive<OpenSHA_Module>> {
     // contents here
}

public class FaultSystemSolution extends ModuleContainer<OpenSHA_Module> implements ArchivableModule,
SubModule<ModuleArchive<OpenSHA_Module>> {
     // contents here
}
```

You load a rupture set or solution via static `load(File)` methods, and you write them via `getArchive().write(File)` methods. If you load an rupture set that has a solution attached, the solution will still be written out even if you never accessed the solution (or didn't even know it was there). These classes make all file I/O methods final, such that they can only be loaded and stored in their general form that will always be accessible (e.g., if you make a NZRupSet class it will be loaded back in as a regular FaultSystemRupSet, and thus be accessible in SCEC-VDO). All custom information should be added as modules.

I made the old rupSet/solution implementations extend these new classes, so it's a drop in replacement and we can slowly ween ourselves off of the old classes. This should make adoption less painful.

### File Structure

Here's what it looks like as a zip archive:

* _modules.json_: master module index, lists rupture set and solution modules
* _ruptures_: sub-directory for rupture information & sub-modules
  * _ruptures/modules.json_: rupture set sub-modules
  * _ruptures/section_indices.csv_: data file used by the rupture set module
  * _ruptures/properties.csv_: data file used by the rupture set module
  * _ruptures/info.txt_: data file used by the Info module that was added to the rupture set module
* _solution_: sub-directory for solution information & sub-modules
  * _solution/modules.json_: solution sub-modules
  * _solution/rates.csv_: data file used by the solution module
  * _solution/info.txt_: data file used by the Info module that was added to the solution module

...and here are the contents of those various module.json files:

modules.json:

```
[
  {
    "name": "Rupture Set",
    "className": "org.opensha.sha.earthquake.faultSysSolution.FaultSystemRupSet",
    "path": "ruptures/"
  },
  {
    "name": "Solution",
    "className": "scratch.UCERF3.FaultSystemSolution",
    "path": "solution/"
  }
]

```

ruptures/modules.json:

```
[
  {
    "name": "Info",
    "className": "org.opensha.sha.earthquake.faultSysSolution.modules.InfoModule",
    "path": "ruptures/"
  }
]
```

solution/modules.json:

```
[
  {
    "name": "Info",
    "className": "org.opensha.sha.earthquake.faultSysSolution.modules.InfoModule",
    "path": "solution/"
  }
]
```

## TODO list

* Deprecate old classes
* Finalize file formats for rupture sets. [Draft documentation here](https://github.com/opensha/opensha/wiki/Modular-Fault-System-Solution).


src/main/java/org/opensha/sha/earthquake/faultSysSolution/ruptures/util/SegmentationCalculator.java
```diff
@@ -1407,8 +1408,8 @@ public static void main(String[] args) throws IOException, DocumentException {
 		ClusterConnectionStrategy connStrat = new InputJumpsOrDistClusterConnectionStrategy(rupSet.getFaultSectionDataList(), distAzCalc, jumpDist, new ArrayList<>());
 		RuptureConnectionSearch rsConnSearch = new RuptureConnectionSearch(rupSet, distAzCalc,
 				1000d, RuptureConnectionSearch.CUMULATIVE_JUMPS_DEFAULT);
-		rupSet.buildClusterRups(rsConnSearch);
-		List<ClusterRupture> rups = rupSet.getClusterRuptures();
+		rupSet.addModule(ClusterRuptures.instance(rupSet, rsConnSearch));
```
chrisbc, line 1411: @kevinmilner I think this is potentially problematic, but first my assumptions:

 - that line 1411-new ==  line 1410-old >>  to build the Cluster Ruptures list using some helper class
 - line 1412-new & line 1411-old copy these objects to the local member variable 'rups'

My key concern is that it's reasonably easy to see what is happening in the old API from the method names and introspection in IDE etc. This is a  good thing & for new users especially, as the domain is being 'explained' through the API.  However, in the new API it less clear, as we seem to rely on the module class names and the module 'concept'. It' almost like we're creating a 'language' with only two or three verbs, which I fear may be rather difficult to understand for casual/new users.

kevinmilner:
Lots of progress to report on. First I have implemented a new rupture set builder, including a method for reproducing a UCERF3 inversion rupture set:

```
FaultSystemRupSet.builder(subSects, rups).forU3Branch(branch);
```

I have tested all of the various arrays and such against the original UCERF3 implementations and verified that they match perfectly to full double precision [JUnit here](https://github.com/opensha/opensha/blob/modular-fault-sys-rup-set2/src/test/java/org/opensha/sha/earthquake/faultSysSolution/RupSetBuilderTests.java). Once I got things to match perfectly, I modified both codes a tiny bit to use FaultSection.getArea(boolean) instead of doing their own calculations, for simplicity. That method has a different multiplication order so it doesn't match the previous implementation perfectly to floating point precision, but I think it's better to use the standard methods.

I also modularized all of the inputs necessary for generating inversion inputs, and verified that the matrices match perfectly without using any of the old classes. The same is true after writing the new files to a (new style) archive, and loading them back in, with the caveat that I had to copy over the input fault section data (due to issue #25, same would occur with the old classes and I/O). Those tests are just in the main method of [UCERF3InversionInputGenerator](https://github.com/opensha/opensha/blob/modular-fault-sys-rup-set2/src/main/java/scratch/UCERF3/inversion/UCERF3InversionInputGenerator.java) for now as they require a lot of memory and aren't well suited for CI.

So I think we're in really good shape. @chrisbc  and @voj, I can show you how to configure inversions using the new code next week. Next steps include:

* Serializing grid source providers
* Improving fault section file format
* Generalizing the inversion setup code to be more flexible for UCERF4 (longer term goal)
* Deprecating the old code, and letting the new code load in the old file formats

kevinmilner:
This is almost ready, and I now have a complete draft of the file formats for core modules (gridded seismicity was the last one).

I have posted a draft of file format documentation, which is intended mostly for outside users wishing to implement the models in other codes (it doesn't currently describe the module JSON format). As an outside-ish user, would you be willing to take a look @pmpowers-usgs?

Docs here: https://github.com/opensha/opensha/wiki/Modular-Fault-System-Solution

Demo files here: https://github.com/opensha/opensha/tree/modular-fault-sys-rup-set2/src/test/resources/org/opensha/sha/earthquake/faultSysSolution

I also renamed the old (in scratch) classes to U3FaultSystemRupSet and U3FaultSystemSolution, and deleted some unused one-off test classes.

The updated TODO list is:

* Add formal tests for rupture set/solution file formats, loading, default modules, etc
* Deprecate old classes
* Finalize file formats for rupture sets
* Add fallback code if you try to load the new objects from an old file

pmpowers-usgs:
Yes, but not for a few days

kevinmilner:
I updated the file format following feedback from @chrisbc to separate out the section indices for each rupture, and properties (like magnitude/rake) into separate CSV files. [Documentation has been updated here to reflect these changes](https://github.com/opensha/opensha/wiki/Modular-Fault-System-Solution).

pmpowers-usgs:
Review comments for modular fault system solution:

**Documentation looks great!**

Solution comments:
- `README` --> markdown
- (nit) Don't put the trailing slash in `"path"` in `modules.json` files. Use `java.nio.file` internally to resolve paths.
- Reduce precision across the board on export of the solutions and fix floating point errors where present in magnitudes and lat-lons.

Ruptures comments:
- `section_indices.csv` --> `rupture_indices.csv` | `indices.csv`
  - Drop the `# 1,# 2, ...` part of the header in section_indices; the readme should explain what the additional columns are after the header
- `ave_slips.csv` --> `slips.csv` | `average_slips.csv`
  - Why isn't `slip` in `properties.csv`?
- `fault_sections.geojson` features need `DipDirection`
- `properties.csv` should include dip and width; reduce precision:
  - magnitude to 3 decimal places
  - rake to 0 or 1 decimal place; integer value is fine too
  - `area` and `length` to, say, 5 decimal places (6 sig figs); Is there a compelling reason to have area and length in `m` as opposed to `km`?

Solution comments:
- `grid_region.geojson`
  - I'd be wary of storing grid coordinates as a `MultiPoint` nested inside a `GeometryCollection` (I don't like these because there is no way to define properties for the component geometries). Consider WUS or AK at 0.1° spacing? 4x increase if we go to 0.05°.
  - I don't understand `LatNodes` and `LonNodes`
  - Consider using `grid_region.geojson` as follows:
    - a single `Polygon` feature that is the bounds of the grid
    - include `spacing` property; are lat and lon spacing ever different? To date, NSHMs are always the same and just have a single value
    - consider dropping `anchor`; the stored coordinates should suffice
    - use optional `bbox` geojson member to define mercator bounds of region

- `*_mfds.csv`
  - Fix floating point precision errors in magnitude headers
  - Consider also having `mfds.csv` that user knows are total gridded mfds
  - Reduce precision of mfd rates to 5 or 6 decimal places.


Good rounding method to clean up values to fixed precision when serializing:
``` java
  /**
   * Round a double to a specified number of decimal places according to the
   * supplied {@link RoundingMode}. Internally this method uses the scaling and
   * rounding capabilities of {@link BigDecimal}. Note that a negative
   * {@code scale} will round {@code value} to the specified number of places
   * above the decimal.
   *
   * @param value to round
   * @param scale the number of decimal places in the result
   */
  public static double round(double value, int scale, RoundingMode mode) {
    return BigDecimal.valueOf(value).setScale(scale, mode).doubleValue();
  }
```

kevinmilner:
Thanks a log @pmpowers-usgs, this is really helpful. A few comments/questions/ramblings that come to mind as I mull your suggestions over:

> Reduce precision across the board on export of the solutions and fix floating point errors where present in magnitudes and lat-lons.

I certainly agree that things would be more user friendly with more appropriate precision. The downside of doing so is that [de]serialized files wouldn't exactly match the rupture sets/solutions that were created in memory. So if you created a rupture set in memory and generated inputs for an inversion, those inputs would be slightly different if you had loaded in the same rupture set from a file. Any differences would likely be immaterial, however. I suppose we could do the rounding on construction rather than on serialization to keep things consistent.

> Why isn't slip in properties.csv?

I have gone back and forth on this one. The current thinking is that "average slip" is an optional property that is not required for hazard calculations, just for running inversions (satisfying slip rate constraints), and thus shouldn't be in the required `properties.csv` file. Externally built rupture sets might not have this property defined, and I don't bother with slips when I load in RSQSim catalogs as rupture sets (though the information is there in RSQSim).

> fault_sections.geojson features need DipDirection

Agreed, I actually ran into this yesterday and forced it to always be written from here on out.

> properties.csv should include dip and width

Interesting -- those things are calculated on the fly from the fault sections in OpenSHA, do you think it would be more convenient for users to have them included in the file? In that case I would write them out, but probably ignore them on load. I guess the same could be said for area and length, though lengths aren't always the sum of all subsection lengths (e.g., faults with multiple subsections down-dip).

> I'd be wary of storing grid coordinates as a MultiPoint nested inside a GeometryCollection

Do you propose a different way of storing them? It's very important to me to store the grid nodes rather than infer them, as I don't want to ever be in the position where a change in the Location class or gridding algorithms breaks compatibility with previous results (as occurred recently when we changed Location). I need to reliably be able to load in the exact same grid.

> I don't understand LatNodes and LonNodes

They're optional and can be included to ensure that any deserialized GriddedRegions are exactly match the original version, but can be ignored by most users.

> include spacing property; are lat and lon spacing ever different? To date, NSHMs are always the same and just have a single value

I'm mirroring the GriddedRegions class that supports different spacings, and don't see the harm. I can see it being used for extreme latitudes. You might consider it for Alaska where equal spacing in degrees would mean approximately double the horizontal resolution: at lat=65 N, 0.1 degrees of longitude is 5km whereas 0.1 degrees of latitude is still 11 km.

> use optional bbox geojson member to define mercator bounds of region

Can you elaborate on the usefulness of this?

pmpowers-usgs:
Good points. Further thoughts:

> dip and width

Most users aren't messing with individual ruptures. However, reconstructing average dips and widths from fault sections requires a bunch of steps, some of which I could see users getting wrong (e.g. the width when averaging is measured in the DipDir, not normal to the upper trace of a potentially trapezoidal subsection). I'd gotten used to having them available because they feed directly into GMMs, no muss, no fuss. However, I don't think we (the NSHMP) are parameterizing ruptures the way we should be given how the GMMs are developed and their sensitivity to things like hanging wall effects. Averaged values from multi-fault ruptures give rise to parameterizations that don't make sense or are inconsistent. I'd advocate continuing to export them for now with possible removal/deprecation at a later date.

> grid_region.geojson

As far as I can tell, you've got the actual border of the region encoded in `Lat|LonNodes` and the the rectangular (mercator) 'bounding box' (hence the `bbox` suggestion) encoded in the GC > Polygon feature. The nodes themselves exist in 4 files. In the NSHMs, we use geojson polygons for borders and other properties of the region, but that file is secondary to csv files which enumerate the specific nodes along with MFD properties or explicit rates in magnitude columns. I'd like to know where Ned is on all the bookkeeping business; the on-fault, off-fault, supra-, sub-seimogenic distinctions are a lot of complexity that  we don't _actually_ know is an improvement, or not, over dealing with the just the fault (and fault-like structures, or zones) in the inversion and then computing a final off-fault/background MFD. We're refactoring the NSHMs to use spatial PDFs in the model files. These can be used to scale a-values or an explicit MFD. Double-counting in grid cells then gets handled as the model loads, truncating any node MFD at the mMin of fault-based ruptures in the cell. (In UCERF3 there was then a step where we went back and renormalized the spatial PDF based on the truncations; very confusing and difficult to explain/remember). The only point of this ramble is that it would be much simpler to define solutions that contain a spatial PDF and the background MFD.

kevinmilner:
I think the time has finally come to merge this in, as development on master has ceased, things seem to be in working order, and file formats are now relatively stable. Will also merge in related branches on dev, cybershake, and SCEC-VDO repos.

Issue #25, opened 2021-06-11
Re-examine Location class and degree/radian conversions


kevinmilner:
Our [Location class](https://github.com/opensha/opensha/blob/master/src/main/java/org/opensha/commons/geo/Location.java) stores values internally in radians. I think this was done by @pmpowers-usgs in order to speed up location calculations. I noticed some strange behavior, however, where the following fails due to rounding errors:

```
double lat = 37.09084333164281;
Preconditions.checkState(lat == new Location(lat, 0d).getLatitude());
```

This fails because the latitude gets converted to radians, then getLatitude() returns a version converted back from radians, and that returned value is `37.090843331642816` (notice the extra digit, a `6` at the end).

Internally, the input latitude of `37.09084333164281` becomes `0.6473573384785501`, but when that is converted back to degrees it becomes `37.090843331642816`, which in turn will convert back to radians as `0.6473573384785503`.

There are multiple ways to convert between radians and degrees, and Java actually switched their method in JDK 9+ [thanks to this report](https://bugs.openjdk.java.net/browse/JDK-8051808).

Here's how our particular number is affected by the various algorithms:

```
INPUT DEG	37.09084333164281
Using Java 9+ Math.toRadians/toDegrees:
	RAD:	0.6473573384785501
	DEG:	37.090843331642816
Using Pre-Java 9: /180*PI and *180/PI
	RAD:	0.64735733847855
	DEG:	37.09084333164281
Using BigDecimal
	RAD:	0.6473573384785501
	DEG:	37.09084333164281
```

So the difference for this value is due to the change in Java 9+. But I believe them that the change was for the better, and is more accurate more of the time. I tried their sample code and found that, for 1000 trials, the Java 9 method wins 2420 of the time, Pre-Java 9 wins 336 times, and the rest are ties. So this particular latitude is probably one of those rare cases where the old code worked better.

I see a few options:

1. Go back to storing things in degrees, as that's how they're almost always created/serialized
  a. PROS: Serialization will always work, equals() will always work as expected
  b. CONS: LocationUtils calculations will be slightly slower, but I'm not sure by how much
2. Store both degrees and radians
  a. PROS: Serialization will always work, equals() will always work as expected, calculations will still be as fast
  b. CONS: Memory requirements of Location class increase (must store 2 more double values), and this is a core class that can exist many thousands of times in memory
3. Switch to BigDecimal for accurate conversion, still just store in radians
[...]

PR #27, opened 2021-06-17
Location class now stores both degrees and radians internally

kevinmilner:
This is in response to #25, that notes that serialization is broken in the prior implementation that converts to radians and only stores that version internally.

The proposed modification uses 56 bytes per instance, a 40% increase over the previous version, but remains as efficient for distance and related calculations. For 1 million location instances in memory, that results in a 15 MB increase in memory requirements, whereas distance calculations without pre-computed radian values take ~30% longer (and are the primary hazard calculation bottleneck, so many calculations would slow down by this amount).

I also explicitly made the values public final, and use the raw double values in equals/hashCode() implementations (also making getL*itude() final).


kevinmilner:
The unit tests are failing after this change, which brings up 2 interesting and somewhat scary points. With those points in mind, I'm hesitant to act on this right now, and might just live with the serialization issues for now. I welcome outside opinions though!

## GridReader issues

The FSS/U3 ERF tests are failing, to first order, because [GridReader](https://github.com/opensha/opensha/blob/location-serialization/src/main/java/scratch/UCERF3/griddedSeismicity/GridReader.java) has a NaN value when loading the UCERF2 spatial seismicity PDF. After some digging, I found that the new location class very slightly changes the gridding of the RELM testing region....yikes:

The old version has 7636 grid nodes, and the first 3 are:

```
31.60000, -117.20000, 0.00000
31.60000, -117.10000, 0.00000
31.60000, -117.00000, 0.00000
```

The new version has 1 extra, 7637 grid nodes, and the first 3 are:

```
31.60000, -117.30000, 0.00000
31.60000, -117.20000, 0.00000
31.60000, -117.10000, 0.00000
```

This makes me nervous about making this change, as there could be other similar issues out there that aren't caught by current tests, and I'm not quite sure how to fix this instance in a good way. We relied on consistent gridding of those regions in UCERF3, and should design UCERF4 to be insensitive to these sorts of things details.

## Location serialization and servlets

The other tests (site data) are failing because they talk to a servlet that currently uses the old version of the Location class, and thus java serialization fails. Making this change will require that anyone using our servlets also update their code, which could get tricky at the next release. This won't be a huge issue on the master branch and nightly builds, but anyone who has previously downloaded a release will get exceptions if they try to access site data with old jars once a new version of the applications is released with this change. The error message won't be intuitive, and I don't think it's something we can really intercept server side to make things happier.

kevinmilner:
Upon further thought, we need to address this now. Serialization is going to be very important, especially with respect to gridded seismicity in UCERF4/NZ models.

I added GeoJSON serialization support for our Region classes (Region and GriddedRegion), and now the CaliforniaRegions class loads in GeoJSON representations that were created with the old Location. This means that, even after making this change, the CaliforniaRegions class will still return the exact same regions with the exact same gridding as was used in UCERF3.

I had to implement one small hack to make this work: GridSourceFileReader now detects the issue when trying to load the UCERF3 gridded region from XML (as is the case for fault system solution zip files), and loads the new JSON version instead. This only applies to the old FSS zip files, and won't be used for any new models going forward (or once we migrate all of the UCERF3 data files to the new format).

Finally, I was able to make the operational site data tests pass by making the accessors use double arrays instead of Location instances. This option was originally added with backwards compatibility in mind, allowing anyone to use the servlets even with a different Location instance. Unfortunately, the currently released applications don't use that option so we'll have some issues when the release branch switches to the new Location, but we can defer that for some time.

I think this is ready to be merged in, and plan to do so in the next day or so without objection.

pmpowers-usgs:
Can you get rid of Java serialization?

On the NSHM side, locations are always serialized as part of a GeoJSON coordinate array or csv. The Location object itself is never serialized.

kevinmilner:
> Can you get rid of Java serialization?
> 
> On the NSHM side, locations are always serialized as part of a GeoJSON coordinate array or csv. The Location object itself is never serialized.

The java serialization is only used for servlets, and switching to GeoJSON would still break compatibility with our currently released applications, but it could be worth doing at the next release. The servlet might be more useful as a GeoJSON servlet (e.g., it could be used by other codes), but that would require more effort. I'll think about that next time we make a formal release.

mbarall:
For the record, I have seen cases where the new Location class causes the earthquake lat/lon fetched from ComCat to be slightly different than it was with the old Location class.

kevinmilner:
@mbarall yeah that's not surprising. If you need to match old values for any reason, the new Location.backwardsCompatible(lat,lon.depth) method should do the trick.

PR #33, opened 2021-06-30
fix for trace buffers of faults that turn back on themselves

voj:
Trace buffers for faults that turn back on themselves do not always contain the whole fault trace. E.g.:
image (1)

Because the (red/green) fault curves back against itself in the south, the (blue) buffer polygon is constructed in a faulty way.

This fix will construct the following polygon instead:

image

### DIFF
--- a/src/main/java/scratch/UCERF3/griddedSeismicity/SectionPolygons.java
+++ b/src/main/java/scratch/UCERF3/griddedSeismicity/SectionPolygons.java
@@ -168,9 +168,9 @@ class SectionPolygons {
 
 			// merge buf and fault polys if buf > 0 (i.e. dip > 45)
 			double scaledBuf = scale * buf;
-			LocationList buffPoly = buildBufferPoly(trace, dipDir, scaledBuf);
+			Area buffArea = buildBufferPoly(trace, dipDir, scaledBuf);
 			Area pArea = parentAreaMap.get(pID);
-			pArea = merge(pArea, new Area(buffPoly.toPath()));
+			pArea = merge(pArea, buffArea);
 			pArea = cleanBorder(pArea);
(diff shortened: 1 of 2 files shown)


kevinmilner:
Great catch @voj. Merging it in will be a bit tricky, however, as it changes some UCERF3 polygons and will break compatibility with the published model. Here's a comparison of the the polygons with your change to the original polygons, for both UCERF3 fault models:

FM3.1:
292/2606 have a different number of points
Of those that have the same number of points:
	380/2314 are not exactly equal (double precision)
	189/2314 are not approximately equal (float precision)

FM3.2:
298/2664 have a different number of points
Of those that have the same number of points:
	394/2366 are not exactly equal (double precision)
	200/2366 are not approximately equal (float precision)

I think the path forward will be to serialize the UCERF3 polygons so that they remain forever unchanged before merging this in. The recent change to Location (see #27) also affects these polygons to a lesser degree (I controlled for this in my comparisons above), so I need to serialize them anyway. Let me work on that.

voj:
@kevinmilner and here I was thinking that this would be a trivial change 😅 Thanks for being on top of things.

kevinmilner (approved):

kevinmilner:
Looks good @voj. I serialized all of the UCERF3 polygons, now available on the modular branch and will be used going forward. I also visually inspected the differences between the new approach (after your fix) and the original polygons, and found no visible difference (besides the points being slightly different). You can view that comparison at the following link for each fault section where the polygon wasn't identical. The original polygons are drawn with thick blue lines, and the new ones thinner red lines that match perfectly to the eye.

https://gist.github.com/kevinmilner/e2e05f280ec8404d6479353b0166d412

PR #34, opened 2021-07-06
make gridSourceProvider protected

chrisbc:
So we can over-ride getter/setter gridSourceProvider methods as needed in regional subclasses of FaultSystemSolution.


kevinmilner:
This is going to get messy with the modular branch as the gridSourceProvider object no longer exists and the old FaultSystemRupSet has been renamed. Can you instead just call setGridSourceProvider in your subclass constructor to avoid the U3-specific logic in InversionFaultSystemSolution for now?

chrisbc:
Hi Kevin, I considered that approach but because we need the gridSP for
subduction models to be null (at least for now). So it won't work because
later when getGridSourceProvider is called the current U3 logic will kick
in and set up the U3 provider. I had thought that this would be
straightforward, but if not we can look for another alternative. Just leave
it for now, we've made that change in our fork and it's working OK and we
can discuss it next week if needed.

kevinmilner:
Closing this as it's not compatible with the new to-be-merged-in modular approach

PR #41, opened 2021-08-20
fixes #40: shutdown ExecutorService

### LINKED ISSUE #40: JVM does not terminate after using ThreadedSimulatedAnnealing
`ThreadedSimulatedAnnealing` creates an `ExecutorService` which needs to be shut down manually in order for the JVM to terminate.

voj:
closes #40 

I noticed the `ExecutorService` is not simply local to the `iterate` method, not sure what the intention is. I've simply added a `shutdown` method to the class, although I think it's safer to simply always shutdown the service at the end of the `iterate` method.

Alternatively, you might want to consider making the TSA `Closeable`, so that we can use the try-with-resources statement https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

### DIFF
--- a/src/main/java/scratch/UCERF3/simulatedAnnealing/ThreadedSimulatedAnnealing.java
+++ b/src/main/java/scratch/UCERF3/simulatedAnnealing/ThreadedSimulatedAnnealing.java
@@ -505,6 +505,16 @@ public class ThreadedSimulatedAnnealing implements SimulatedAnnealing {
 		long[] ret = { iter, perturbs };
 		return ret;
 	}
+
+	/**
+	 * Shuts down the thread pool.
+	 */
+	public void shutdown(){
+		if (exec != null){
+			exec.shutdown();
+			exec = null;
+		}
+	}
 	
 	public static String timeStr(long millis) {
 		double secs = millis/1000d;


kevinmilner:
Good point! I set it up that way so that, when there's an averaging layer on top, the intermediate layer of `ThreadedSimulatingAnnealing` workers don't constantly re-create thread pools.

2 other approaches to d this would be:

1. Provide a custom `ThreadFactory` instance that supplies daemon threads instead, [see this SO answer](https://stackoverflow.com/a/39221278)
2. Make `ThreadedSimulatingAnnealing` close thread pools at the end of the `iterate(...)` method by default, but have a method for disabling that behavior so that higher level `ThreadedSimulatingAnnealing` instances can signal to their own workers that they should re-use their pools

I think that I lean toward option 2 as it should be transparent to the user. I'll give it a try, but comments welcome.

voj:
Method 2 sounds simpler at first, but then you'd also have to come up with a way of shutting down the threads when you're done averaging. Sounds like daemon threads would handle both cases automatically.

kevinmilner (approved):

kevinmilner:
OK I'll implement the daemon threads so that it never holds things up when we forget to shutdown, but also merge this in so that we can explicitly clean things up as well

PR #45, opened 2021-09-02
added new MFD Uncertainty weighted constraint; 

chrisbc:
This PR create two new classes to support the concept of MFD constraints having weights by bin:
 - **MFD_WeightedInversionConstraint** which extends MFD_InversionConstraint, adding the `EvenlyDiscretizedFunc weights;` field
 - **MFDUncertaintyWeightedInversionConstraint** which used the above to assign uncertainty weights

It also adds a new method  to EvenlyDiscretizedFunc -  **setYofX** which expects a user defined function (e.g. a lambda function) …  @kevinmilner I did this because so much code using this class does this mapping outside the class, straight after creating an instance.

### DIFF
--- a/src/main/java/org/opensha/commons/data/function/EvenlyDiscretizedFunc.java
+++ b/src/main/java/org/opensha/commons/data/function/EvenlyDiscretizedFunc.java
@@ -208,7 +208,7 @@ public class EvenlyDiscretizedFunc extends AbstractDiscretizedFunc{
 
 	/**
 	 * Returns the minimum y-value in this series. Since the value could
-	 * appear aywhere along the x-axis, each point needs to be
+	 * appear anywhere along the x-axis, each point needs to be
 	 * examined, lookup is slower the larger the dataset. <p>
 	 *
 	 * Note: An alternative would be to check for the min value every time a
@@ -224,7 +224,7 @@ public class EvenlyDiscretizedFunc extends AbstractDiscretizedFunc{
 
 	/**
 	 * Returns the maximum y-value in this series. Since the value could
-	 * appear aywhere along the x-axis, each point needs to be
+	 * appear anywhere along the x-axis, each point needs to be
 	 * examined, lookup is slower the larger the dataset. <p>
 	 *
 	 * Note: An alternative would be to check for the min value every time a
@@ -240,7 +240,7 @@ public class EvenlyDiscretizedFunc extends AbstractDiscretizedFunc{
 	
 	/**
 	 * Returns the x index for the maximum y-value in this series. Since the value could
-	 * appear aywhere along the x-axis, each point needs to be
+	 * appear anywhere along the x-axis, each point needs to be
 	 * examined, lookup is slower the larger the dataset. <p>
 	 *
 	 */
--- a/src/main/java/org/opensha/sha/earthquake/faultSysSolution/inversion/constraints/impl/MFDUncertaintyWeightedInversionConstraint.java
+++ b/src/main/java/org/opensha/sha/earthquake/faultSysSolution/inversion/constraints/impl/MFDUncertaintyWeightedInversionConstraint.java
@@ -1,0 +1,104 @@
+package org.opensha.sha.earthquake.faultSysSolution.inversion.constraints.impl;
+
+import java.util.HashSet;
+import java.util.List;
+
+import org.opensha.commons.data.function.EvenlyDiscretizedFunc;
+import org.opensha.sha.earthquake.faultSysSolution.FaultSystemRupSet;
+import org.opensha.sha.earthquake.faultSysSolution.inversion.constraints.InversionConstraint;
+import org.opensha.sha.magdist.IncrementalMagFreqDist;
+
+import com.google.common.base.Preconditions;
+
+import cern.colt.matrix.tdouble.DoubleMatrix2D;
+import scratch.UCERF3.utils.MFD_InversionConstraint;
+import scratch.UCERF3.utils.MFD_WeightedInversionConstraint;
+
+/**
+ * Constrain the solution to match the given MFD Weighted constraints. 
+ * 
+ * @author chrisbc
+ *
+ */
+public class MFDUncertaintyWeightedInversionConstraint extends InversionConstraint {
+	
+	public static final String NAME = "MFD UncertaintyWeighted";
+	public static final String SHORT_NAME = "MFDUncertaintyWeighted";
+	
+	private FaultSystemRupSet rupSet;
+	private double weight;
+	private List<MFD_WeightedInversionConstraint> mfdWeightedConstraints;
+	private HashSet<Integer> excludeRupIndexes;
+
+	public MFDUncertaintyWeightedInversionConstraint(FaultSystemRupSet rupSet, double weight,
+			List<MFD_WeightedInversionConstraint> mfdWeightedConstraints) {
+		this.rupSet = rupSet;
+		this.weight = weight;
+		this.mfdWeightedConstraints = mfdWeightedConstraints;
+	}
+
+	@Override
+	public String getShortName() {
+		return SHORT_NAME;
+	}
+
+	@Override
+	public String getName() {
+		return NAME;
+	}
+
+	@Override
+	public int getNumRows() {
+		return MFDEqualityInversionConstraint.getNumRows(mfdWeightedConstraints, rupSet);
+	}
+
+	@Override
+	public boolean isInequality() {
+		return false;
+	}
+
+	@Override
+	public long encode(DoubleMatrix2D A, double[] d, int startRow) {
+		long numNonZeroElements = 0;
+		// Loop over all MFD constraints in different regions
+		int numRuptures = rupSet.getNumRuptures();
+		for (int i=0; i < mfdWeightedConstraints.size(); i++) {
+			double[] fractRupsInside = rupSet.getFractRupsInsideRegion(mfdWeightedConstraints.get(i).getRegion(), false);
+			IncrementalMagFreqDist targetMagFreqDist = mfdWeightedConstraints.get(i).getMagFreqDist();
+			EvenlyDiscretizedFunc targetUncertaintyWeight = mfdWeightedConstraints.get(i).getWeights();
+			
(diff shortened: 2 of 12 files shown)


chrisbc:
@kevinmilner @voj I'm not quite ready to merge just want to do some more real-world testing, but I'm just letting you know this is in the pipeline. Do let me know if there's anything here needs work :)

src/main/java/org/opensha/commons/data/function/EvenlyDiscretizedFunc.java
```diff
@@ -339,6 +341,29 @@ public void set(double x, double y) {
...
+	 * @param mappedFn(x)
+	 */
+	public void setYofX(DoubleUnaryOperator mappedFn){
+		for(int i=0; i<this.size();i++) {
+			this.set(i, mappedFn.applyAsDouble(getX(i)));
+		}		
+	}	
+	
+	/**
+	 * Maps a user-defined function F to each point, setting y = F(x).
```
voj, line 356: should be `y = F(x , y)`

src/main/java/org/opensha/commons/data/function/EvenlyDiscretizedFunc.java
```diff
@@ -339,6 +341,29 @@ public void set(double x, double y) {
 		points[index] = y;
 	}
 
+	/**
+	 * Maps a user-defined function F to each point, setting y = F(x).
+	 * 
+	 * @param mappedFn(x)
+	 */
+	public void setYofX(DoubleUnaryOperator mappedFn){
```
voj, line 349: You could easily put this as high in the hierarchy as `AbstractXY_DataSet`

src/main/java/org/opensha/sha/earthquake/faultSysSolution/inversion/constraints/impl/MFDUncertaintyWeightedInversionConstraint.java
```diff
@@ -0,0 +1,111 @@
...
+		return SHORT_NAME;
+	}
+
+	@Override
+	public String getName() {
+		return NAME;
+	}
+
+	@Override
+	public int getNumRows() {
```
voj, line 51: looks equal to the implementation in `MFDEqualityInversionConstraint` apart from the constraints list name. Maybe pull this out into a shared method?

src/main/java/org/opensha/commons/data/function/EvenlyDiscretizedFunc.java
chrisbc: @voj Ahh, yes of course. For now I might leave it here and see what Kevins thoughts are about that.

src/main/java/org/opensha/sha/earthquake/faultSysSolution/inversion/constraints/impl/MFDUncertaintyWeightedInversionConstraint.java
chrisbc: @voj And also the same as MFDInequality... I agree this might be refactored to be more DRY, but I'll let Kevin review as-is first.

chrisbc:
@kevinmilner  FYI this branch was rebased on the fix https://github.com/opensha/opensha/pull/45/commits/66bd2313d60ce98015153d489cfacd76ef3a0a14 as I need both running. I should probably have create a new branch instead - my bad.

src/main/java/scratch/UCERF3/inversion/InversionTargetMFDs.java
```diff
@@ -155,7 +155,7 @@ private Precomputed() {
 		public Precomputed(InversionTargetMFDs targetMFDs) {
 			this(targetMFDs.rupSet, targetMFDs.getTotalRegionalMFD(), targetMFDs.getTotalOnFaultSupraSeisMFD(),
 					targetMFDs.getTotalOnFaultSubSeisMFD(), targetMFDs.getTrulyOffFaultMFD(),
-					targetMFDs.getMFD_Constraints(), targetMFDs.getOnFaultSubSeisMFDs());
+					(List<MFD_InversionConstraint>) targetMFDs.getMFD_Constraints(), targetMFDs.getOnFaultSubSeisMFDs());
```
kevinmilner, line 158: In concert with my suggested adapter change, you could make the other `Precomputed` constructor take a `List<? extends MFD_InversionConstraint>` and remove this cast

src/main/java/scratch/UCERF3/inversion/UCERF3InversionConfiguration.java
```diff
@@ -260,7 +260,7 @@ public static UCERF3InversionConfiguration forModel(InversionModels model, Fault
 		double parkfieldConstraintWt = 1000;
 		
 		// get MFD constraints
-		List<MFD_InversionConstraint> mfdConstraints = targetMFDs.getMFD_Constraints();
+		List<MFD_InversionConstraint> mfdConstraints = (List<MFD_InversionConstraint>) targetMFDs.getMFD_Constraints();
```
kevinmilner, line 263: any reason not to make the method a `List<? extends MFD_InversionConstraint>`?

src/main/java/org/opensha/sha/earthquake/faultSysSolution/inversion/constraints/impl/MFDUncertaintyWeightedInversionConstraint.java
kevinmilner: Looks fine to me, but I agree that this method could be moved to a shared static method. I went ahead an added one in 59d418d that you can use. Otherwise looks good, with the caveat that we may end up renaming these things down the line if we go with a different uncertainty scheme

src/main/java/org/opensha/commons/data/function/EvenlyDiscretizedFunc.java
kevinmilner: I like @voj's suggestion here, and may even end up using this stuff. I say add it to the XY_DataSet interface as a default implementation instead (only putting it in the abstract class means it won't work on plain `DiscretizedFunc` references, which are common).

kevinmilner (changes requested):
A few minor changes in my comments, but it generally looks good to me. We may end up renaming things in the future as we figure out how we plan to use MFD constraints in UCERF4, but fine for now

chrisbc:
@voj @kevinmilner thanks both for the feedback & suggestions. I'll attempt to turn this around today - cheers!

src/main/java/org/opensha/sha/earthquake/faultSysSolution/inversion/constraints/impl/MFDUncertaintyWeightedInversionConstraint.java
voj: done

src/main/java/scratch/UCERF3/utils/MFD_WeightedInversionConstraint.java
```diff
@@ -0,0 +1,115 @@
...
+		return weights;
+	}
+	
+	@Deprecated
+	@Override
+	public Element toXMLMetadata(Element root) {
+		throw new UnsupportedOperationException("No more XML, sorry");
+	}
+	
+	public static class Adapter extends TypeAdapter<MFD_WeightedInversionConstraint> {
```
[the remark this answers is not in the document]
voj: Hmmm, you've got a good point. I think this is where the general JSON approach of not adding type tags to objects falls over. In EDN for example, you'd register your adapters with the parser and you'd just tell it to "read the next object" and it would do the right thing. Your suggestion is a bit like manually implementing duck-typing while also breaking the inheritance hierarchy. That said, it makes things work better and I don't have a better idea. 🤔

src/main/java/scratch/UCERF3/inversion/InversionTargetMFDs.java
voj: done

src/main/java/scratch/UCERF3/inversion/UCERF3InversionConfiguration.java
voj: I don't think there was a special reason for it

src/main/java/scratch/UCERF3/utils/MFD_WeightedInversionConstraint.java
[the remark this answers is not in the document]
voj: Done

src/main/java/scratch/UCERF3/utils/MFD_InversionConstraint.java
```diff
@@ -132,6 +133,11 @@ public void write(JsonWriter out, MFD_InversionConstraint value) throws IOExcept
 				Region region = new Region(value.region);
 				regionAdapter.write(out, region);
 			}
+
+			if(value instanceof MFD_WeightedInversionConstraint){
```
chrisbc, line 137: @voj @kevinmilner  I think it's not great to have a parent class be 'aware' of it's subclass(es). Ideally the subclass would override the parent class and mutate the json returned from calling the superclass method, but that's not  possible with this API (where the method returns void).

src/main/java/scratch/UCERF3/utils/MFD_WeightedInversionConstraint.java
```diff
@@ -0,0 +1,115 @@
...
+	
+	
+	public EvenlyDiscretizedFunc getWeights() {
+		return weights;
+	}
+	
+	@Deprecated
+	@Override
+	public Element toXMLMetadata(Element root) {
+		throw new UnsupportedOperationException("No more XML, sorry");
```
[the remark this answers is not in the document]
chrisbc: :) heh

src/main/java/scratch/UCERF3/utils/MFD_InversionConstraint.java
voj: @chrisbc I agree that the broken class hierarchy is not great. We could move the Adapter class into its own file and then that would be fixed.

voj:
@kevinmilner we're done with our changes

kevinmilner (approved):

PR #50, opened 2021-09-10
Added filled polygons to plots 

voj:
Adds (optional) solid fill to surfaces in `RupSetMapMaker`, with `SlipRatePlots` making use of it.

This is to make subduction plots easier to read. I haven't touched the geojson code, I can make it match if you prefer.

I added a polygon option to `PlotLineType` but didn't change the name to something like `PlotFunType` because I didn't want to change too much.

Here's an example of how filled surface polygons can make outliers easier to spot for a subduction plot:

image (3)

slip_rates_sol_ratio (2)

### DIFF
--- a/src/main/java/org/opensha/commons/gui/plot/PlotLineType.java
+++ b/src/main/java/org/opensha/commons/gui/plot/PlotLineType.java
@@ -5,11 +5,7 @@ import java.awt.Shape;
 import java.awt.Stroke;
 import java.util.NoSuchElementException;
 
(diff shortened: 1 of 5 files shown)


voj:
@kevinmilner oops, just realised I'd left the commit message as "WIP". It's not a WIP :-)

kevinmilner (approved):
Very cool, especially how easy it was to add the fill capability in PlotLineType. I'll definitely use this elsewhere when making plots. I can go ahead and do the GeoJSON part myself.

PR #51, opened 2021-10-11
deleting redundant average slip implementations

voj:
Removed two implementations of average slip from InversionFaultSystemRupSet so that the single source of truth (at least for the implementation) is in the AveSlipModule


kevinmilner:
This PR reminds me that you're still using InversionFaultSystemRupSet, and I advise you to ween yourself off of it! I had been leaving as much as possible as is in there in order to be able to test that we could still reproduce UCERF3 (e.g., org.opensha.sha.earthquake.faultSysSolution.RupSetBuilderTests compares a rupture set built with IFSRS and the new methods). But that reproducibility has been well demonstrated, so I'm ok merging this in and removing the duplicated code.

PR #52, opened 2021-10-11
Allowing LogicTreeBranches to have nodes with custom JSON serialization

voj:
This PR lets nodes provide their own JSON TypeAdapter. This allows us to use complex nodes, such as scaling relationships with multiple config values.

Deserialization where the level and adapter class are unknown (e.g. SCEC VDO) is handled gracefully.

### DIFF
--- a/src/main/java/org/opensha/commons/logicTree/JsonAdapterHelper.java
+++ b/src/main/java/org/opensha/commons/logicTree/JsonAdapterHelper.java
@@ -1,0 +1,111 @@
+package org.opensha.commons.logicTree;
+
+import com.google.common.base.Preconditions;
+import com.google.gson.TypeAdapter;
+import com.google.gson.annotations.JsonAdapter;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonWriter;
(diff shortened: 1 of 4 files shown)


src/main/java/org/opensha/commons/logicTree/JsonAdapterHelper.java
```diff
@@ -0,0 +1,111 @@
+package org.opensha.commons.logicTree;
```
kevinmilner, line 1: This is a useful utility class that we might promote to a more general location and reuse, but here is fine for now

src/main/java/org/opensha/commons/logicTree/LogicTreeLevel.java
```diff
@@ -259,7 +306,7 @@ public void write(JsonWriter out, LogicTreeLevel<? extends E> level) throws IOEx
 				case "enumClass":
 					enumClassName = in.nextString();
 					break;
-				case "className":
+				case "class":
```
kevinmilner, line 309: thanks for catching and fixing my bug here!

PR #65, opened 2021-11-30
fixed #64: passing on rake to TMG_CRU_2017 and TMG_SUB_2017

### LINKED ISSUE #64: ScalingRelationShips: TMG_CRU_2017 and TMG_SUB_2017: rake must be set when getting aveSlip and Mag
`ScalingRelationShips.TMG_CRU_2017` and `ScalingRelationShips.TMG_SUB_2017` must both pass the rake to their MagArealRel with `setRake()` when calculating aveSlip and Mag.

voj:
closes #64


kevinmilner:
I'm a bit surprised that you're still using the old UCERF3 ScalingRelationships interface, is there a reason you're using it rather than your own RupSetScalingRelationship implementation?

voj:
We've implemented new scaling relationships in our repo, but some of our rupture sets still refer to this one. So serialisation would be easier if this one was fixed.

kevinmilner:
OK, I'll merge but please ween yourself when you can!

Issue #66, opened 2021-12-01
Uncaught exceptions in SectBySectDetailPlots leading to 'phantom' thread execution


chrisbc:
When an IllegalStateException is raised by a sub-plot within the SectBySect loop, the exception is recast as a runtime error and thrown. In our NZSHM use case calling thread catch will the exception and terminate, leaving the SectBySect thread(s) to continue running .

Ideally these exceptions should be reported to stderr and the process should continue attempting to produce sub-reports.

kevinmilner:
How about you try again after this fix (4b0cda4), does that do the trick? I think I want the full plot to fail if an exception is encountered, but to still continue on at the report level to create any other plots. What exceptions are you getting? I'd rather just fix them then skip faults that trigger some edge case.

chrisbc:
Thanks @kevinmilner

PR #67, opened 2021-12-01
Fix for 66 phantom threads after rethrown exception

chrisbc:
closing #66 - we report the issue and continue processing.


chrisbc:
@kevinmilner I see you've already tackled this , so we have a merge conflict. In our current case we prefer to produce the majority of the sub-reports (95%) that do not raise exceptions as this is giving the users valuable science. Meanwhiel we're tackling the root problem in another ticket on our downstream repo (see https://github.com/GNS-Science/nzshm-opensha/issues/88)

kevinmilner:
@chrisbc OK, I incorporated your changes in 783da0f with the modification that skipped faults will still show up in the table, just without a link. I wanted to avoid things just disappearing from the table without being noticed, so keeping them there will still make it clear when there's a problem. 1b875d7 will also probably fix your b-value calc exception.

chrisbc:
awesome - thanks @kevinmilner

PR #69, opened 2021-12-02
Fix for #68, exception in  sect b value plot when sectMinMax > maxMagForSect

### LINKED ISSUE #68: SectBValuePlot can't handle a max mag that is less than min mag
It is possible for a section to have a smaller max mag than the min mag. This is because min mag may be constrained by a min min mag while max mag is not.

SectBValuePlot can't handle this case and throws an exception. https://github.com/opensha/opensha/commit/1b875d783915b3aec3bb4723677203f2f9003e02 was supposed to remedy this, but now a different exception is triggered:

```
no point at index -1
java.lang.IndexOutOfBoundsException: no point at index -1
	at org.opensha.commons.data.function.EvenlyDiscretizedFunc.getX(EvenlyDiscretizedFunc.java:276)
	at org.opensha.sha.earthquake.faultSysSolution.reports.plots.SectBValuePlot$BValEstimate.<init>(SectBValuePlot.java:354)
	at org.opensha.sha.earthquake.faultSysSolution.reports.plots.SectBValuePlot.estBValue(SectBValuePlot.java:322)
```

Here is a test to reproduce the Exception https://github.com/opensha/opensha/commit/96b8e0680964735f5d8f31240bd900f8c65d38f0

I'm not sure what to do. Make the plot handle this gracefully? Make `FaultSystemRupSet.getMaxMagForSection()` observe the min min mag?

kevinmilner:
Closes #68

### DIFF
--- a/src/main/java/org/opensha/sha/earthquake/faultSysSolution/reports/plots/SectBValuePlot.java
+++ b/src/main/java/org/opensha/sha/earthquake/faultSysSolution/reports/plots/SectBValuePlot.java
@@ -351,8 +351,8 @@ public class SectBValuePlot extends AbstractSolutionPlot {
 				}
 			}
 			this.numBins = 1+maxIndex-minIndex;
-			this.minMagBin = refFunc.getX(minIndex);
-			this.maxMagBin = refFunc.getX(maxIndex);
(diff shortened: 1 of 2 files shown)


voj:
@kevinmilner I was able to successfully create a report with this fix. Thank you.

PR #71, opened 2021-12-08
[#70] ModuleContainer: fix loadAllAvailableModules

### LINKED ISSUE #70: ModuleContainer: addition order of available modules should not matter
Available modules can depend on other available modules. Currently, if module A depends on module B, then B needs to have been made available before module A. 

Otherwise, `ModuleContainer.loadAllAvailableModules` will trigger the precondition at the beginning of `loadAvailableModule` because it will manually try to load module B that has already been loaded transitively by loading module A.

voj:
fixes #70


kevinmilner (approved):

kevinmilner:
Good catch!

PR #75, opened 2021-12-15
Remove deprecated  RupSetDiagnosticsPageGen

kevinmilner:
RupSetDiagnosticsPageGen was the predecessor to all of the plots in `org.opensha.sha.earthquake.faultSysSolution.reports`. It is deprecated, and contains a fair amount of now unused duplicate code. This pull request deletes it.

@chrisbc @voj can you confirm that you're no longer using it at all? safe to delete?


voj:
@kevinmilner as far as I can tell, we only use it in obsolete tools. Some other uses we can easily fix. @chrisbc could you please confirm?

voj:
@kevinmilner we're all good, you can safely merge this from our perspective.

PR #76, opened 2021-12-21
revert renaming of RateDistributionPlot plots

voj:
RateDistributionPlot plots were renamed in 0b95517fb170b1a80aecca74f33d47d8787f586e causing problems at our end. We've got a web app that displays plots etc to the scientists and renaming plots makes things tricky.


voj:
as discussed, no longer required

PR #77, opened 2022-01-13
Significant simulated annealing efficiency improvements

kevinmilner:
cc @chrisbc @voj

This pull request greatly improves efficiency of the simulated annealing algorithm by a factor of ~6-40x. The actual improvements depends on problem size, with the greatest benefit for larger problems (more constraint rows). Before, inversion speed scaled approximately inverse-linearly with row count, but now the dependence on row count is minimal (performance is similar for large and small problems).

Speedup examples:
* Large model with a full nucleation MFD constraint for each subsection
  * 48147 rows
  * before: ~8k iterations/sec
  * after: ~325k iterations/sec
  * speedup: ~40x
* Smaller model with a single nucleation rate constraint for each subsection
  * 5542 rows
  * before: ~60k iterations/sec
  * after: ~370k iterations/sec
  * speedup: ~6x

I think that I have now (really this time) squeezed just about every ounce of possible performance out of the algorithm. Details on the changes are below.

## Removed all Arrays.copyOf(...) and System.arraycopy(...) invocations from the iteration loop

This was the primary bottleneck for large problems. Every iteration had 1 System.arraycopy(...) on misfit vectors (2 if there are inequality constraints). Additional array copies were possible in certain worse case scenarios as well, e.g., randomly keeping non-ideal perturbations. Now, I keep track of misfits for the best, current kept model, and current perturbed model (that we may or may not keep depending on energy and temperature) separately. Instead of doing full array copies, I copy (or roll back) only rows that were modified in those misfit arrays, as needed.

I also removed an array copy on the solution on every kept worse perturbation.

These had an enormous impact on older compute nodes at CARC that presumably don't have as much memory bandwidth.

## Moved conditional statements out of core loops

I also moved some if statements out of the core energy calculation loops. This leads to slightly more duplicated code, but the duplicated code is simple and side by side.

## Minor performance improvements to sum of squares calculations

I now use `Math.fma(a, b, c)` to do sum of squares calculations, which both reduces floating point error and can be more efficient (can be a single floating point operation).

I also now 'unroll' the sum of squares loop which helps the JVM to vectorize it, leading to ~15% speedup in my tests. See [here](https://stackoverflow.com/a/51429748) and [here](https://en.wikipedia.org/wiki/Loop_unrolling) for more information.

## Now calculate energy change in each iteration, rather than recalculating full energy

This is the big one for small problems that were already pretty efficient.

After those previous optimizations are applied, the bulk of the remaining time is spent on energy calculation (summing the square of all misfits), which scales with row count. We need not recalculate the sum of squares for every row in the misfit array, however, as most are unchanged in any iteration. Instead, now when calculating misfits I keep track of energy change:

```
deltaE += new_misfit*new_misfit - old_misfit*old_misfit
```

Then I add that energy difference to the previously computed energy. Due to floating point precision issues, this new energy calculation method will differ slightly from the old version where we summed all rows, but I think it is actually likely more accurate. Before, if energy is large (as is the case early on), small individual constraint misfits could be ignored. For example, if energy is 1e8 and you try to add a misfit of size 1e-9, that added misfit is ignored to double precision:

```
jshell> 1e8+1e-9
$3 ==> 1.0E8
```

So, it should be more accurate to sum up the smaller energy changes, then add that smaller number to the large number (prior energy).

I recalculate the energy in full every 100k iterations, or at the end of each single thread loop, to reduce any slow accumulation of floating point errors by summing many differences. This leads to fractional errors (relative to the full energy calculation) on the order of 1e-14.

## Other minor changes

These improvements require knowing what rows in the inputs each column (rupture) affects. We were already using this information previously in the misfit calculation step if the input matrix was column compressed, but now we need it more places. I now use a new `ColumnOrganizedAnnealingData` instance that manages this information and is created by (or passed in to) the constructors of the `SimulatedAnnealing` implementations. If an input A matrix is not already column compressed, it will be here.

The `SimulatedAnnealing` interface (and thus the threaded and serial implementations) now return a new immutable `InversionState` object with more details than just the total iteration count. That object is also now used in the `CompletionCriteria` interface. This was done to make tracking inversions easier, and also to facilitate adding additional information that may be needed by any `CompletionCriteria` implementations without having to further modify the method signature (just add it to the state object instead).

New `IterationsPerVariableCompletionCriteria` where the number of iterations is dependent on the number of variables we're solving for (rupture count). For example, if you use this as a sub-completion criteria with a value of '1', then across-thread synchronization will haven every numRuptures iterations (meaning that each rupture has been randomly selected once on average). Or if you set it very large, you could use it as a full completion criteria to, for example, end an inversion after each rupture has been selected an average of 1000 times. I'm planning on using this rather than time steps in the future as it should adapt better to varying problem sizes (small or large rupture sets) and won't behave differently for slower running models (e.g., with more constraint rows).

The number of worse perturbations randomly kept is now tracked by the algorithm, stored in the AnnealingProgress module, and plotted as a function of iteration count.

My dynamically-reweighted inversion implementation is now in here and working, see `org.opensha.sha.earthquake.faultSysSolution.inversion.sa.ReweightEvenFitSimulatedAnnealing`. It will also track constraint misfit progress (beyond energy) with the new `InversionMisfitProgress` module and plots.

Fixed an unrelated issue with `ModuleContainer` where adding an available module didn't evict any matching already loaded modules. This wasn't necessarily a bug, but was counter intuitive and led to bugs in user code. Now, if you add an available module, any already loaded modules that match the given type will be evicted.

### DIFF
--- a/src/main/java/org/opensha/commons/hpc/pbs/USC_HPCC_ScriptWriter.java
+++ b/src/main/java/org/opensha/commons/hpc/pbs/USC_HPCC_ScriptWriter.java
@@ -4,7 +4,7 @@ import java.io.File;
 import java.util.ArrayList;
 import java.util.List;
 
-
+@Deprecated
 public class USC_HPCC_ScriptWriter extends BatchScriptWriter {
 	
 //	public static final File MPJ_HOME = new File("/home/rcf-12/kmilner/mpj-v0_38");
--- a/src/main/java/org/opensha/commons/util/modules/ModuleContainer.java
+++ b/src/main/java/org/opensha/commons/util/modules/ModuleContainer.java
@@ -322,15 +322,40 @@ public class ModuleContainer<E extends OpenSHA_Module> {
 	/**
 	 * Adds an available module that will be lazily loaded when {@link ModuleContainer#getModule(Class)} or
 	 * {@link ModuleContainer#hasModule(Class)} is called and no module is presently loaded that matches the given class.
+	 * <p>
+	 * This will also remove any already loaded modules that map to the given class, otherwise this available module
+	 * could be shadowed. If you want to attach an available module anyway even if a match is already loaded, use
+	 * {@link #addAvailableModule(Callable, Class, boolean)}, but note that the available module will only ever be loaded
+	 * if the the loaded module is first removed. If you simply want to offer an available module that is only kept if
+	 * no match yet exists, such as a default implementation, use {@link #offerAvailableModule(Callable, Class)}.
 	 * 
 	 * @param call
 	 * @param moduleClass
 	 */
 	@SuppressWarnings("unchecked")
 	public synchronized <M extends E> void addAvailableModule(Callable<? extends OpenSHA_Module> call, Class<M> moduleClass) {
+		addAvailableModule(call, moduleClass, true);
+	}
+	
+	/**
+	 * Adds an available module that will be lazily loaded when {@link ModuleContainer#getModule(Class)} or
+	 * {@link ModuleContainer#hasModule(Class)} is called and no module is presently loaded that matches the given class.
+	 * 
+	 * @param call
+	 * @param moduleClass
+	 * @param removeMatchingLoaded if true, any already loaded classes that map to the given class will be removed,
+	 * otherwise this module will not replace any already loaded modules.
+	 */
+	@SuppressWarnings("unchecked")
+	public synchronized <M extends E> void addAvailableModule(Callable<? extends OpenSHA_Module> call,
+			Class<M> moduleClass, boolean removeMatchingLoaded) {
 		List<Class<? extends OpenSHA_Module>> assignableClasses = getAssignableClasses(moduleClass);
 		
 		// fully remove any duplicate associations
+		if (removeMatchingLoaded)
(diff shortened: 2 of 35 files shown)


kevinmilner:
Well, I lied...there was more performance to be squeezed out of the algorithm...a lot more!

I was initially disappointed when I tried the new optimized code on the USC cluster and found that it was still pretty slow on my large problem, only a factor of 2 better than before the optimizations. So I did a really deep dive into the algorithm to find the bottlenecks, and realized that (a) VisualVM sometimes lies about where exactly a thread is spending its time, and (b) there was one more array copy in there that was the real culprit, especially for large models. I was able to remove that last array copy, and things are now lightning fast and don't really depend much on problem size.

For example, my large model (~50k rows) gets 700k iterations/sec on my workstation, and 325k iterations/sec on the USC cluster. Before any of these optimizations those numbers were 15k and 8k, respectively.

For comparison, the much smaller model (~5k rows) now gets 800k iterations/sec on my workstation and 370k iterations/sec at USC, only a slight improvement over the large model.

I have described the improvements in detail in the now-updated original comment, give it a re-read if you've already taken a look. Enjoy the new speed @voj and @chrisbc, I'll merge in after our call next week.

PR #78, opened 2022-01-20
Fault and Deformation model interfaces + U3 refactoring

kevinmilner:
Extracted interfaces for fault and deformation model capabilities that the old UCERF3 FM/DM enums now implement. The RuptureSets utility class now works with these interfaces rather than their U3-specific instances.

I also removed the U3-specifc builder methods on `FaultSystemRupSet`, with their contents moved to `U3InversionConfigFactory`.

@chrisbc @voj, a couple changes that might impact you:

* `RupSetScalingRelationship` now extends `LogicTreeNode` in order to facilitate getting it from logic tree branches in factory classes. If you have any instances that aren't also a `LogicTreeNode` then you'll need to add the additional branch weight method, which can return 0 or NaN if not used in a logic tree branch
* Refactored some of the UCERF3-specific FSRS/FSS subclasses to prefix them with 'U3', e.g., `U3SlipAlongRuptureSet`. If you're still using these, please ween yourself off of them ASAP
* Deleted the long-deprecated old wrapper class scratch.UCERF3.InversionTargetMFDs class that was kept for backwards compatibility when we renamed it to U3InversionTargetMFDs

I'll wait to merge until I know how much these changes impact you and I've done some additional tests/cleanup on my end.


voj:
@kevinmilner thank for checking, Kevin. It only took me a few minutes to make our repo build again 👍

PR #81, opened 2022-01-28
Feature/uncertainty changes

voj:
Hi Kevin,

These are some fixes/changes I made while adding uncertainties to some of our constraints.

The changes to `SlipRateInversionConstraint` are a little odd in that they make the constructor work as its signature suggests, but at the same time this can create a constraint that behaves differently after deserialization. So not sure what you think of this.

For us, this is a convenient way to slip in a synthesized `SectSlipRates` module with our modified standard deviations. And we don't serialise constraints at the moment. What's the use case for constraint serialisation/deserialisation?


src/main/java/org/opensha/sha/earthquake/faultSysSolution/inversion/constraints/impl/MFDInversionConstraint.java
```diff
@@ -57,7 +57,7 @@ public MFDInversionConstraint(FaultSystemRupSet rupSet, double weight, boolean i
 		this.mfds = mfds;
 		if (weightingType == ConstraintWeightingType.NORMALIZED_BY_UNCERTAINTY) {
 			for (IncrementalMagFreqDist mfd : mfds)
-				Preconditions.checkNotNull(mfd instanceof UncertainIncrMagFreqDist,
+				Preconditions.checkArgument(mfd instanceof UncertainIncrMagFreqDist,
```
kevinmilner, line 60: thanks for catching!

kevinmilner (approved):
Looks good to me

kevinmilner:
Looks good to me, @voj. The main use case for serialization is that I often configure inversions locally and the push the JSON files to the cluster for running, and also for reproducibility down the line. I made those modules transient to avoid duplicating the information, but it might we worth serializing at least the slip rates, which is similar to what we do in the MFD constraint. I'll think about that

PR #94, opened 2022-10-11
NSHM point source object

pmpowers-usgs:
Port of nshmp-lib `PointSourceFinite` object as `PointSourceNshm`. As far as I can tell there is only one significant difference between this and `PointSource13b`. Constructor for `PointSourceNshm` doesn not need `depths[]` array.

### DIFF
--- a/src/main/java/org/opensha/sha/earthquake/rupForecastImpl/PointSourceNshm.java
+++ b/src/main/java/org/opensha/sha/earthquake/rupForecastImpl/PointSourceNshm.java
@@ -1,0 +1,581 @@
+package org.opensha.sha.earthquake.rupForecastImpl;
+
+import static java.lang.Math.ceil;
+import static java.lang.Math.cos;
+import static java.lang.Math.min;
+import static java.lang.Math.pow;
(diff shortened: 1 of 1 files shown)


kevinmilner:
@pmpowers-usgs can you summarize what that significant difference was between this and `PointSource13b`? Also, I assume that NSHM18 used this version, but did NSHM14 use this version or `PointSource13b`? I want to make sure to label things correctly in the code, and probably keep `PointSource13b` as an option in `BackgroundRupType` if it's what was used in NSHM14.

pmpowers-usgs:
Yes, the same point source model was used for 2014 and 2018 (our UCERF3 implementation didn't change). The new `NshmPointSource` has some refactoring in it that will make it a little more flexible if you want to use it for CEUS point sources down the road. Right now it (and `PointSource13b`) are essentially hard coded with the WUS depth distribution model. _The only significant difference between the two is the removal of the minimum 0.5 rJB distance from `PointSource13b`_. In `NshmPointSource`, if the site is coincident with the point source location, rJB = 0.0.

PR #100, opened 2023-10-20
Update gradle wrapper to 8.4

kevinmilner:
This is needed for more recent versions of eclipse and more recent java versions. It now supports Java 21


kevinmilner:
Note that everyone will have to do a right click on project -> Gradle -> Refresh Gradle Project on each project after this update (along with pulling updates to each project).

PR #103, opened 2024-03-18
Updated cutoffs

### LINKED ISSUE #102: Update TectonicRegionType distance cutoffs
Hazard calculations performed with OpenSHA should adhere to the max applicable distances prescribed by the GMMs. One way to achieve this is through the use of the `TectonicRegionType.defaultCutoffDist()`. However, the max distance cutoff (the maximum site-to-source distance considered in the hazard integral) values are wrong relative to modern GMMs.

pmpowers-usgs:
Resolves #102


kevinmilner:
Sounds good. These aren't actually used anywhere currently, but will be a good basis for adding TRT-dependent cutoff distances.

PR #105, opened 2024-04-04
SectIDRange memory usage improvements

voj:
## Problem
The particular rupture set that I'm currently investigating has ca 860 fault sub sections and takes ca 40 minutes to build with a 50GB heap which is nearly all used up towards the end.

Looking at the memory usage at 33 minutes, the largest chunk by far is taken up by `SectIDRange` instances:
- 15GB by `short[]` which looks related to `ShortIDRange` since the number of objects is nearly identical (473,958,109 vs 473,958,103)
- 11GB by `SectIDRange:ShortIDRange` 
- 2.5GB by `SectIDRange:SingleID`

Screenshot 2024-04-04 160926

## Approach
Modifications to `SectIDRange`:
- replaced arrays with primitive instance variables
- added `ShortSingleID`
- 473,958,103 individual `ShortIDRange` instances are way too many for 860 sub sections, so I've added caching to ensure that we do not keep multiple identical ranges in memory. For this application, a simple hash map is sufficient.

## Result
The rupture build now finishes after 24 minutes. It must have been getting too close to the 50GB limit before, resulting in the usual Java slowdown. Here is a `VisualVM` screenshot at 20 minutes, showing that the three problem areas `SectIDRange:ShortIDRange`, `SectIDRange:SingleID`, and `short[]` are now no longer pertinent. `short[]` no longer features, the other two are at the bottom of this screenshot.

Screenshot 2024-04-04 215910

### DIFF
--- a/src/main/java/org/opensha/sha/earthquake/faultSysSolution/ruptures/util/SectIDRange.java
+++ b/src/main/java/org/opensha/sha/earthquake/faultSysSolution/ruptures/util/SectIDRange.java
@@ -2,6 +2,9 @@ package org.opensha.sha.earthquake.faultSysSolution.ruptures.util;
 
 import com.google.common.base.Preconditions;
 
+import java.util.HashMap;
+import java.util.Map;
+
 /**
  * Range of section IDs, inclusive, for memory efficient tracking of unique ruptures and contains
  * operations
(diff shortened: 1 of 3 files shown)


src/main/java/org/opensha/sha/earthquake/faultSysSolution/ruptures/util/SectIDRange.java
```diff
@@ -10,53 +13,70 @@
  *
  */
 public abstract class SectIDRange implements Comparable<SectIDRange> {
-	
+
+	private final static Map<SectIDRange, SectIDRange> cache = new HashMap<>();
```
kevinmilner, line 17: Should this be a ConcurrentHashMap since build() isn't synchronized?

voj: Good call. I've changed it. Doesn't seem to impact performance.

kevinmilner (approved):

kevinmilner:
Looks, good, thanks and keep em coming :)

PR #106, opened 2024-04-10
Make cacheSs2R optional

voj:
This PR changes the `AggregatedStiffnessCalculator.CACHE_SS2R` constant to an instance variable that can be configured in order to save heap space.

I'm experimenting with creating rupture sets that contain both crustal and subduction faults. We cannot create larger sets at the moment because our heap is flooded with `SectIDRange` arrays, 20GB in this screenshot:

Screenshot 2024-04-06 205339

Note how the number of instance matches the number of `UniqueRupture` objects.

I found that setting `AggregatedStiffnessCalculator.CACHE_SS2R` to `false` will completely eliminate this problem with no impact on the running time. A further check showed that our builds actually never have a positive hit for this cache.

### DIFF
--- a/src/main/java/org/opensha/sha/simulators/stiffness/AggregatedStiffnessCalculator.java
+++ b/src/main/java/org/opensha/sha/simulators/stiffness/AggregatedStiffnessCalculator.java
@@ -89,6 +89,7 @@ public class AggregatedStiffnessCalculator {
 		}
 		CACHE_ARRAY_SIZE = maxTerminalIndex+1;
 	}
+	private final boolean cacheSs2R;
 	
 	// multiplier to use for section IDs, to which patch IDs are added to get unique patch IDs
 	private transient int patch_sect_id_multiplier = -1;
@@ -539,6 +540,7 @@ public class AggregatedStiffnessCalculator {
 		private SubSectStiffnessCalculator calc;
 		private List<AggregationMethod> layers;
 		private boolean allowSectToSelf = false;
+		private boolean cacheSs2R = true;
 
 		private Builder(StiffnessType type, SubSectStiffnessCalculator calc) {
 			this.type = type;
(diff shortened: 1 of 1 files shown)


kevinmilner:
Another good catch! I can confirm that for NSHM23, performance is better with it disabled:

```
CACHE_SS2R = true
Built 553,116 ruptures in 346.71 secs = 5.78 mins. Total rate: 1,595 rups/s
CACHE_SS2R = false
Built 553,116 ruptures in 211.48 secs = 3.52 mins. Total rate: 2,615 rups/s
```

I didn't look at memory consumption, but it makes sense that it would be less. I'm tempted to just globally disable it (set CACHE_SS2R = false) rather than merge this in with the option, since I don't really see us going back, but could be convinced otherwise. Thoughts?

voj:
> Another good catch! I can confirm that for NSHM23, performance is better with it disabled:
> 
> ```
> CACHE_SS2R = true
> Built 553,116 ruptures in 346.71 secs = 5.78 mins. Total rate: 1,595 rups/s
> CACHE_SS2R = false
> Built 553,116 ruptures in 211.48 secs = 3.52 mins. Total rate: 2,615 rups/s
> ```
> 
> I didn't look at memory consumption, but it makes sense that it would be less. I'm tempted to just globally disable it (set CACHE_SS2R = false) rather than merge this in with the option, since I don't really see us going back, but could be convinced otherwise. Thoughts?

I'm very happy to have it always disabled if it improves both our performance and yours. I didn't know if it relied on our particular configuration so I wanted to play it safe. I've updated the PR to do this now.

If you saw a speed up with the cache disabled, I suspect that you hadn't assign a large enough heap to the JVM. There's a gray-area between having enough memory and exiting with an out of memory exception, were the JVM can just make do with what it has but the garbage collection churn is so high that you experience a noticeable slow-down. If you're unlucky your process can nearly grind to a halt without ever telling you that something is wrong.

Here's a [VisualVM](https://visualvm.github.io/index.html) example where the used heap (blue in the right graph) is converging on a relatively narrow band that looks like its safely below the maximum heap size. This will eventually lead to a situation where the CPU cycles are mostly used for garbage collection. (this example is actually a rupture building process with `CACHE_SS2R` set to true)
Screenshot 2024-03-14 111742

kevinmilner:
Increasing heap (to 180 GB, which the process never reached) did bridge the gap a little bit (down to 4.6m for my problem), but still didn't reach the no-cache performance. There must be some overhead in the caching process that's just not worth it; the lower level patch-to-patch stiffness calculations are already cached, this was caching some of the aggregate measures. I remember testing these back in the day, but something else must have changed that made the cache no longer useful (maybe on of your earlier updates?).

Thanks again!

PR #107, opened 2024-04-15
New SourceFilter interface and implementations

kevinmilner:
This introduces a new `SourceFilter` interface and adds a tectonic-region specific distance filter option. A new GUI has been created for selecting and modifying the various filter options. The default implementation remains unchanged for now (fixed distance of 200 km).

`SourceFilter`s can filter at the source and/or rupture level. The tectonic region distance filter only applies to sources as ruptures don't currently know their tectonic region.

Because event sets are created by the ERF, I added a new `ERF.drawRandomEventSet(Site, Collection<SourceFilter>)` method that respects these filters. HazardCurveCalculator uses this new method.

Finally, this updated implementation can introduce one very minor change to hazard curve calculations for sources where the source surface is larger than individual rupture surfaces. Previously, in full probabilistic calculations (but not in event set calculations), all rupture of a source would be included if the source surface is within the distance cutoff. That was true even if individual ruptures were further than the distance cutoff. The new implementation checks distances at both the source and rupture level, which means that event set calculations will now use the same filtering scheme as full probabilistic. But a few ruptures that may have been included in earlier full probabilistic calculations may now be excluded.

### DIFF
--- a/src/main/java/org/opensha/commons/gui/LabeledBorderPanel.java
+++ b/src/main/java/org/opensha/commons/gui/LabeledBorderPanel.java
@@ -126,7 +126,10 @@ public class LabeledBorderPanel extends JPanel{
 	 */
 	public void setTitle( String newTitle ) {
 		title = newTitle;
-		if( titledBorder1 != null ) titledBorder1.setTitle( title );
+		if( titledBorder1 != null ) {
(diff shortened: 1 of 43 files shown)


kevinmilner:
Here's what the new GUI looks like:

image

Issue #109, opened 2024-04-16
Refactor GUIs and calculators to avoid using unsafe Thread.stop() to cancel calculations


kevinmilner:
If you start a calculation in the GUIs, you can cancel it via the cancel button, which causes Thread.stop() on the calculation thread. That method has been long deprecated by Java and is slated for removal. It is unsafe because it can leave objects in an unexpected state.

Implementations should instead use the interrupt mechanism to notify threads that they should stop. The tricky part is that then calculations (or updateForecast() implementations) would need to check for interrupts, adding programming overhead. I don't think it's feasible to update ERFs to check for interrupts, but it won't be too hard to add to the calculators. So I propose that we:

* remove all Thread.stop() references
* check for interrupts after updating the forecast (a cancel will not succeed until after updateForecast() completes). 
* check for interrupts after every source in the hazard curve calculator

kevinmilner:
@abhatthal, can you look into this? I think this can be implemented as part of a larger refactor of threading in our applications. I noticed Java has indeed deprecated it to the extent that calling `stop()` will throw an exception in newer releases, so we need to deal with this before the next release.

Currently, when someone hits compute in, e.g., `HazardCurveApplication`, this is the flow:

```java
calcThread = new Thread(this);
calcThread.start();
```

That calcThread runs the `run()` method of `HazardCurveApplication` in a separate thread, which calls `computeHazardCurve()`. We can probably move to using `CompletableFuture` instances instead. I don't love the current method of the application itself implementing `Runnable` and creating a `new Thread(this)`.

`computeHazardCurve()` calls `erfGuiBean.getSelectedERF()`, which in turn calls `erf.updateForecast()`, then it does the actual hazard curve calculation. If a user cancels, it could be detected between updateForecast() and the cancellation.

We could also add some sort of signalling behavior to `HazardCurveCalculator` to be able to detect a cancellation inside the the source loop, but we would need to make sure that checking for a cancellation is instantaneous as that code is the the inner loop of a lot a big calculations. __That is very important: we can't slow this code down.__ The simplest might be to add an `AtomicBoolean` to `HazardCurveCalculator` that represents a canceled state with a setter method, and to check if it has been set inside the source loop (probably not the inner rupture loop) of `computeHazardCurve` and exit if so. I think there are other mechanisms of thread signaling that might be worth looking into (but I don't have any experience with them).

All of these changes will also need to be applied to `HazardSpectrumApplication`, but you can start with `HazardCurveApplication`.

[...]

PR #110, opened 2024-04-16
Java deprecation fixes

kevinmilner:
This deals with a number of deprecation warnings that occur when you try to build OpenSHA with a recent version of java, e.g. Java 21.

* Replace all boxed primitive constructors (e.g., new Double(0)) with valueOf static references
* Applets are dead; refactor out references to Applet classes (replace with JPanel or JFrame)
* Removed references to a few ancient exceptions

I'll make similar changes to the dev and cybershake project directly (without a pull request).

Note that this does not fix the warnings related to the unsafe reliance of the apps on Thread.stop() to cancel calculations. Fixing that will take a bit more work, and I've created a ticket for that here: #109


mbarall:
Regarding replacement of boxed primitive constructors, I have already made that change in opensha-oaf.

PR #112, opened 2024-05-01
OutputCSVFile for writing large CSV files.

voj:
We've started to create large rupture sets while exploring joint ruptures. This is mostly due to the large number of sections in our subduction faults.

This PR introduces a `CSVWriter` and a `CSVReader` that allow us to write the generated ruptures to file without requiring tens of gigabytes of heap. This is mainly achieved by not populating the `List<List<E>> values` property of `CSVFile`. The largest `indices.csv` file we have written with this so far was 4.9GB on disk.

### DIFF
--- a/src/main/java/org/opensha/commons/data/CSVReader.java
+++ b/src/main/java/org/opensha/commons/data/CSVReader.java
@@ -1,175 +1,89 @@
 package org.opensha.commons.data;
 
-import com.google.common.base.Preconditions;
-
 import java.io.*;
 import java.util.*;
 
 /**
- * Memory efficient CSV reader. Not thread safe.
- * Calling next() advances the reader to the next row.
- * get*() methods can be used to get values from the columns of the current row.
- * next() must have been called at least once before a call to a get*() method.
+ * CSV reader that allows for CSV files to be read directly instead of
+ * having to load a potentially large intermediate structure into memory.
  */
 public class CSVReader implements Closeable {
 
-    protected LargeLineBuffer buffer;
+    protected BufferedReader reader;
 
-    // the data of the current row
-    protected List<String> line = null;
-
-    public CSVReader(InputStream in) throws IOException {
-        buffer = LargeLineBuffer.fromStream(in);
+    public CSVReader(InputStream in) {
+        reader = new BufferedReader(new InputStreamReader(in));
     }
 
     /**
-     * Advances the reader to the next row.
+     * Reads the next row.
      *
-     * @return true if the row exists, false if the end of the data has been reached.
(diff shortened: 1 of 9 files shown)


voj:
@kevinmilner hold on with reviewing this, I've got new ideas :-)

kevinmilner:
> @kevinmilner hold on with reviewing this, I've got new ideas :-)

No problem @voj, I'm at a conference this week and won't get to it until next week anyway!

voj:
All ready now!

src/main/java/org/opensha/commons/data/CSVReader.java
kevinmilner, line 1: It looks like CSVReader reads the whole file into memory, did you consider making it just iterate over each line instead? That would require removing the getNumRows() method (and changing FaultSystemRupSet.loadRupSectsCSV(...) to use a while loop), but it should be even more memory efficient and maybe faster. It would also simplify things a bit as I'm finding LargeLineBuffer a bit hard to follow.

src/main/java/org/opensha/commons/data/CSVWriter.java
kevinmilner, line 1: It's not clear to me that having a `CSVWriter` interface (with it's only method being `writeToStream(OutputStream)` is a value add over just having it be a concrete class that does what `CSVWriterStream` currently does. The only method that I see that currently takes the interface rather than a specific implementation is the static `CSV_BackedModule.writeToArchive(...)` method that could just be duplicated which might simplify things? I don't feel strongly about this though, and could also be missing some current and/or future utility of setting it up this way. Mostly just putting this out here to see what you think.

src/main/java/org/opensha/commons/data/CSVReader.java
voj: I had considered that but I wanted to respect the boundary that `CSV_BackedModule` enforces by being the only place that accesses the archive file. It opens and closes the stream and it' very clean. 
However, I have tried it out now and the code is much simpler to read. I've also re-written the reader to return a `Row` object to make it much easier to understand what's going on.

src/main/java/org/opensha/commons/data/CSVWriter.java
voj: After accepting that `CSVReader` could just wrap a zip stream directly, I also changed `CSVReader` to be the simplest `OutputStream` wrapper. Everything is much more straightforward now.

Some of the changes will be easier to see with whitespace hidden 
image

voj:
@kevinmilner I've done a major rework after your comments

kevinmilner (commented):
Looks great, much simpler. Thanks!

kevinmilner (approved):

PR #115, opened 2024-05-07
MagAreaRelationship is not thread safe

voj:
My apologies, https://github.com/opensha/opensha/pull/113 needs to be partially reverted because magnitude calculation is not necessarily thread safe. This is because `MagAreaRelationship` may set the `rake` property before calculating the magnitude:

https://github.com/opensha/opensha/blob/master/src/main/java/org/opensha/commons/calc/magScalingRelations/MagAreaRelationship.java#L60


kevinmilner (approved):
Good catch!

PR #118, opened 2024-05-23
Coulomb subduction playground

voj:
Hi Kevin,

here's a framework for testing out coulomb filters for joint ruptures. I've put it in a more or less random package for now since it's all experimental.

Do you have access to our repo? If not we'll have to work out how we can both work on the same branch.

- There's a `main` method in `RuptureMerger` that opens a rupture set with crustal and subduction ruptures (get it from https://share.gns.cri.nz/123S72ZHG3ZK/rupset-disjointed.zip.html ) and merges ruptures into one particular subduction rupture. There's commented out code to use 50 ruptures instead of just one. But It's rather slow already and creates a lot of ruptures.
- There's a `RuptureJumpDistFilter` that will ensure that ruptures are within 5km of each other.
- A `MultiRuptureCompatibilityFilter` is expected to return a `PlausibilityResult` as well as a `RuptureJump`. Not sure if the jump is important. I guess we could just as well take the first section of each rupture and use those for the jump. But in case we need something more specific, the filter can return it. Any filter that you write won't need to return any jump as the `RuptureJumpDistFilter` already returns one.

Hope this all makes sense and does what you need for now. Not wedded to any of this. Happy to make any changes you need!


src/main/java/org/opensha/sha/earthquake/faultSysSolution/ruptures/MultiRupture/RuptureJump.java
kevinmilner, line 1: Maybe rename this to MultiRuptureJump? All jumps are jumps within ruptures, so the MultiRupture will make it clear what this class means

kevinmilner:
Looks like a great start! I won't have time to try it out until the second half of next week, but it looks good. One thought is that you might rename the package from `MultiRupture` to `multiRupture` as per the usual convention.

voj:
@kevinmilner will you need knowledge about the structure of the subduction ruptures, e.g. a 2DRuptureTreeNavigator?

src/main/java/org/opensha/sha/earthquake/faultSysSolution/ruptures/MultiRupture/RuptureJump.java
voj: done

kevinmilner:
> @kevinmilner will you need knowledge about the structure of the subduction ruptures, e.g. a 2DRuptureTreeNavigator?

Not for the simple filters I'm proposing we start with. We can bite that bullet when/if the time comes (and we hopefully won't have to).

voj:
@kevinmilner here's a small rupture set that can be used during development. It comes with a handy report that helps you choose specific rupture combinations. http://nzshm-sundry-reports.s3-website-ap-southeast-2.amazonaws.com/reports/catalogue-6-June-24/

kevinmilner:
Hi @voj,

I put together a simple `MultiRuptureCompatibilityFilter` implementation that checks that the that at least X % of interactions from one rupture to the other are positive (i.e., slip on one encourages slip on the other). What's the best way for me to contribute it? Should I make my own branch on opensha that forks from your subduction playground then submit a PR to your branch, or do you want to give me access to your branch directly?

A few thoughts that I had while implementing this:

- Not sure that `MultiRuptureCompatibilityFilter.apply(...)` needs to take the previous result as an argument.
- Not sure why `MultiRuptureCompatibilityResult` has a Jump attached, how do you see that being used? Could probably just use `PlausibilityResult`.
- Might eventually want to make `MultiRuptureCompatibilityFilter` extend `ShortNamed` if we end up using multiple of them at a time and want more advanced diagnostics.

Also, do you have the `rupset-disjointed.zip` file referenced in the main method of `RuptureMerger` somewhere? It looks like the rupture set you linked to above might only has a few, and none of them pass my initial Coulomb tests. That's because the 2 test subduction ruptures you included are both either partially or fully down-dip of the crustal faults, which is not Coulomb compatible (but the up-dip patches, 2, 3, and 4, are). Here are some SCEC-VDO screenshots showing this; green are the 'source' patches in the Coulomb calculation, and I'm plotting compatibility where red is compatible and blue is incompatible:

image
image

I did a test calculation where I built my own subduction ruptures using each unique combination of the uppermost 3 ruptures; I got 256 combined ruptures, here's the resulting rupture set:

[mergedRupset.zip](https://github.com/user-attachments/files/15745626/mergedRupset.zip)

voj:
@kevinmilner oh, exciting news! So happy to see you've made progress 🎉 

- Here's a new link to the rupture set. I just learned that it's automatically deleted from the sharing server after 7 days https://share.gns.cri.nz/04XJHYXQR9GN/rupset-disjointed.zip.html
- @chrisbc will give you access to our repo so we should be able to both work on my branch 👍 
- previous result in `MultiRuptureCompatibilityFilter.apply(...)`: OK, I'll remove it
- `MultiRuptureCompatibilityResult`'s `Jump`: I wasn't sure if it's actually important to know where the jump was taken. I'm more than happy to take this out since a single section-to-section jump might not make sense in this context anyway.
- `ShortNamed`: I'll do that too.

That's surprising to me that those ruptures are incompatible 🤔 I was going off some RSQSim earthquakes like these (not by me) from the same area 
image
image

Do you think the incompatibility is due to the specific properties of these ruptures or is there a general incompatibility with the subduction rupture down-dip of the crustal rupture?

I'm far from being a  SCEC-VDO power-user. Is this a built-in feature or debugging code you've added specifically for this?

voj:
@kevinmilner I've pushed changes incorporating your suggestions

kevinmilner:
Here's a Coulomb calculation using a bunch of the subduction sections as sources, showing that some of the crustal faults are encouraged and others aren't:
image

It seems to vary by fault type, maybe liking strike-skip crustal ruptures more. Here's a plot of rake, where red is right-lateral strike-slip, blue is reverse, and yellow is normal:
image

The Coulomb calculates are a poorly documented feature of SCEC-VDO. I just pushed some updates so you should pull first, but when you load a rupture set into VDO you can choose Color By -> Stiffness. You select your first source subsection by ctrl + clicking on it, and then you can add additional ones by shift + ctrl + clicking on them. It can be a bit sluggish, especially for large rupture sets.

Thanks for the updates and giving me access, I'll  merge my initial ideas in shortly!

kevinmilner:
OK I just committed my draft Coulomb filters in `org.opensha.sha.earthquake.faultSysSolution.ruptures.multiRupture.impl`, and updated the main method of `RuptureMerger` to use them.

`MultiRuptureNetCoulombPositiveFilter`: This ensures that the ruptures are net positive in a Coulomb sense, summing all interactions from the source rupture to the receiver rupture and testing that the sum is >0. It can either be specified to require that either crustal -> subduction or subduction -> crustal is positive (`Directionality.EITHER`), or to require that both are positive (`Directionality.BOTH`). This type of filter weights nearby interactions greatest.

`MultiRuptureFractCoulombPositiveFilter`: This ensures that at least `threshold` fraction of interactions between the source rupture and the receiver rupture are positive (regardless of distance or strength of that interaction).

It first does a pre-cache of all coulomb calculations in parallel to speed things up because the rupture building loop is serial as currently implemented when only using a single nucleation rupture. It stores that cache in /tmp and will reuse it. It may take a little while to calculate the first time!

I do find that it still prefers to include corupture where the crustal fault is near the top of the subduction rupture, but that's not always true, e.g.:

image

Here are some test rupture sets using the single subduction rupture and the full set of crustal ruptures:

Fraction interactions >= 0.75: 
[mergedRupset_5km_cff0.75IntsPos_from97653.zip](https://github.com/user-attachments/files/15779538/mergedRupset_5km_cff0.75IntsPos_from97653.zip)

Net Coulomb positive: 
[mergedRupset_5km_cffNetPositiveBOTH_from97653.zip](https://github.com/user-attachments/files/15779540/mergedRupset_5km_cffNetPositiveBOTH_from97653.zip)

Both combined: 
[mergedRupset_5km_cff0.75IntsPos_cffNetPositiveBOTH_from97653.zip](https://github.com/user-attachments/files/15779541/mergedRupset_5km_cff0.75IntsPos_cffNetPositiveBOTH_from97653.zip)

I'm trying to build a full rupture set (using all possible subduction and crustal ruptures), we'll see how it goes.

voj:
@kevinmilner that's awesome! I've been looking at it today. So many things to play with! I've been able to run a rupture build, it really does take a while to build that cache. Let me know how long it takes you to build a cache for everything!
 
 - In SCEC VDO, is there a way to set the focal point so that it's easier to use for NZ? The help document is not reachable, and I've found were the default focal point is hard coded. But I'm hoping for an easier way than hacking that in :-)
 
 - Are the parameters that go into the stiffness calculation unlikely to change? I.e. will we really only have to calculate that cache once?

chrisbc:
>  SCEC VDO, is there a way to set the focal point so that it's easier to use for NZ? The help document is not reachable, and I've found were the default focal point is hard coded. But I'm hoping for an easier way than hacking that in :-)

@voj I think I can help with this .... I have some notes somewhere on using scecVDO with an NZ configuration

kevinmilner:
> @kevinmilner that's awesome! I've been looking at it today. So many things to play with! I've been able to run a rupture build, it really does take a while to build that cache. Let me know how long it takes you to build a cache for everything!

I downsampled to 2km patches and built a full rupture set in about 4.5 hours...it has 26 million ruptures!!! So we'll clearly have to build better filters.

>     * In SCEC VDO, is there a way to set the focal point so that it's easier to use for NZ? The help document is not reachable, and I've found were the default focal point is hard coded. But I'm hoping for an easier way than hacking that in :-)

The simplest way is to go the "Graticule" tab, click the drop down where you see "California" and select "NewZealand" (which I just added in my commits early today). I usually then go to "Political Boundaries" and enable "Oceania" to get NZ (and a bunch more), or you can double click on Oceania and enable NZ specifically.

>     * Are the parameters that go into the stiffness calculation unlikely to change? I.e. will we really only have to calculate that cache once?

We never changed them in CA (except for some early sensitivity tests). It's possible that other parameters would be better for this style of rupture, but I would say we can treat them as unlikely to change for now. Even if we do change them, we're unlikely to change it frequently.

>     * I can't seem to make `Color By -> Stiffness` work. When I select it it's all greyed out, which I assume is what it's supposed to look like. But ctrl+click doesn't on a fault section doesn't seem to have nay effect.

Had you loaded a rupture set first? That's the first step, then click the "Visible" checkmark next to it in the table below the browse button, then you can start coloring and such. All faults will initially be gray when you select stiffness, but you should be able to ctrl+click then.

voj:
@kevinmilner @chrisbc Thanks, the "Graticule" tab was what I needed to change the focal point 👍 

And I finally have the stiffness visualisation working. It seems ctrl+ click doesn't do anything, but shift+ctrl+click does work. Maybe it's a Windows thing?

4.5 hours is amazing! And the result of that is probably the best joint rupture set we have yet 🤩 Is that 4.5 hours including building the stiffness cache? And going to a 1k resolution would take about 4 times as long?

26 million ruptures is a much, much better result than I had thought! The theoretical worst case of rupture combinations at 5km max distance is in the billions. In the meantime, I've looked into thinning. I might be able to give you a rupture set with fewer subduction and crustal ruptures tomorrow. That might bring the result into more manageable territory. With any luck I won't have messed up the fault sections list and the stiffness cache will be reusable.

Oh, this is very exciting!

kevinmilner:
> It seems ctrl+ click doesn't do anything, but shift+ctrl+click does work. Maybe it's a Windows thing?

Oops, I misremembered the key combos! It's shift + click for the first selection, ctrl + shift + click for additional. But with none selected, adding ctrl doesn't do any harm. So that explains it!

The 4.5 hours was for both caching and building the rupture set. When doing the whole set I don't precache as the rupture set is built in parallel anyway and will thus do the caching work in parallel. I would expect that a 1km run would take longer but not 4x. I think 2km spacing is totally defensible for this sort of run though, especially during the prototyping phase.

I just pushed a commit with some tweaks for full calculation. I also posted my [rupture set and 2km cache file here](https://data.opensha.org/ftp/kmilner/misc/2024_nz_crustal_subduction/) in case you want to take a look or rerun reusing my cache. Note that I'm running this on my absolute beast of a machine (64 thread ryzen threadripper 7975) so it might take significantly longer on a laptop even if using my cache file.

voj:
Hi @kevinmilner 
I've had a closer look at the rupture set I gave you. It includes all sections but for some reasons not all of our ruptures. I've created a new rupture set that includes all ruptures ([download it here](https://share.gns.cri.nz/45MLK3TU9F14/nzshm_complete_merged.zip.html)). It'll be better to use this new rupture set. The fault sections are identical apart from the parent ids of the subduction sections. The stiffness cache you created with the 4.5 hour run still works for this rupture set, but it doesn't seem to be quite complete for these rupture combinations.

I have also created lists of rupture ids that can be used to select subsets of this new file. There's new code in `RuptureMerger` that makes use of these lists ([download them here](https://share.gns.cri.nz/2Z3BZERPKIH1/filteredRuptures.zip.html))

- `filteredRuptures.txt` heavily thinned subduction ruptures (ruptures need to be 10% larger than the previous rupture size, ruptures of the same size should not overlap) and crustal ruptures (max cumulative azimuth change 10deg, max cumulative rake change 10deg)
- `filteredRuptures-allCrustal.txt` heavily thinned subduction ruptures, all crustal ruptures
- `filteredRuptures-straightCrustal.txt` heavily thinned subduction ruptures, crustal ruptures have max cum azimuth change of 10deg
- `filteredRuptures-uniformRakeCrustal.txt` heavily thinned subduction ruptures, crustal ruptures have max cum rake change of 10deg

`filteredRuptures.txt` cuts the source ruptures right down. With this file, I can run `RuptureMerger` in just over 2 minutes on my 6-core machine.

voj:
@kevinmilner I've generated a basic rupture report to get some data on how well the approach is working. You can find one that is based on the strongest thinning [here](http://nzshm-sundry-reports.s3-website-ap-southeast-2.amazonaws.com/reports/stiffness-19-June-24/).

The code is on this branch. In order to make it work, I needed to slightly change how we create splays in order to not trip up the `ClusterRuptures` module. Old rupture sets need to be re-generated as they would crash `ClusterRuptures`. 

I've also stuffed the stiffness cache as a module into the rupture set to make sure we use the correct data for the report. We'll have to see how practical this is in the long run. I see this as a tool to use while we experiment.

Here is one [interesting example](http://nzshm-sundry-reports.s3-website-ap-southeast-2.amazonaws.com/reports/stiffness-19-June-24/resources/jointStiffness14023/index.html) of a rupture. It happens to be the rupture with the longest number of sections.

A couple of properties strike me as odd with this one. Not sure if this is due to my ignorance or if there's something there:
- This rupture was accepted because stiffness was above the threshold when using crustal as the source. Does this mean that we're saying that this enormous subduction fault was triggered by the relatively small crustal rupture?
- There's interaction between the crustal rupture and subduction sections in the far north, hundreds of kilometers away. Is this realistic?

I'm thinking we could have a distance-cutoff for section-to-section stiffness. Then we'd also make it impossible for this crustal rupture to trigger the whole subduction fault.

voj:
@kevinmilner to try out my suggestion above ☝️ to limit coulomb at a distance, I've created a new report that explores stiffness with different distance cut-offs http://nzshm-sundry-reports.s3-website-ap-southeast-2.amazonaws.com/reports/stiffness-20-June-24/. Here's the [new version of the interesting example](http://nzshm-sundry-reports.s3-website-ap-southeast-2.amazonaws.com/reports/stiffness-20-June-24/resources/jointStiffness14023/index.html) with a bunch of plots at different distance cut-offs at the bottom (subduction as the source in the top rows, crustal in the bottom rows). Note how this time, none of them reaches the threshold.

kevinmilner:
Great work, @voj! I'm just coming back up to speed.

It looks like you're plotting the fraction of interactions that are positive in your stiffness reports. It all comes down to what summary statistic you take from the set of all patch-to-patch calculation, and [this](https://github.com/GNS-Science/opensha/blob/8d9ee37d9c9b3bec0a9fceb35a45670416c1249a/src/main/java/org/opensha/sha/earthquake/faultSysSolution/ruptures/multiRupture/MultiRuptureStiffnessPlot.java#L47) is where you're setting it up to use the fraction positive.

The fraction positive doesn't take distance into account at all, it just registers if something is positive or negative, so that is why you see such strong far-field compatibility in your example plots. I think we could make the case that we want both crustal -> subduction and subduction -> crustal to be strongly positive, so maybe you should add a `Dicretionality.BOTH` to the call to the `MultiRuptureFractCoulombPositiveFilter` constructor (as I already had in there for `MultiRuptureNetCoulombPositiveFilter`.

Speaking of `MultiRuptureNetCoulombPositiveFilter`, that differs in that it instead sums all Coulomb interactions, thus weighting stronger near-field ones greater than weaker far-field interactions. That filter is supposed to be already doing basically what you were trying to do with the distance cutoff.

I like using a combination of these two filters: are the near-field interactions net positive (`MultiRuptureNetCoulombPositiveFilter`), and is the whole thing also compatible including far-field interactions (`MultiRuptureFractCoulombPositiveFilter`)?

voj:
@kevinmilner yes, I didn't really explain what I was plotting 😅 I set up the `AggregatedStiffnessCalculator` the same way `MultiRuptureFractCoulombPositiveFilter` does to have a closer look at the 0.75 threshold.

I set the directionality to `BOTH` for `MultiRuptureFractCoulombPositiveFilter` as you suggested and ran a [new report](http://nzshm-sundry-reports.s3-website-ap-southeast-2.amazonaws.com/reports/stiffness-21-June-24/). Now the largest rupture by section count is this one http://nzshm-sundry-reports.s3-website-ap-southeast-2.amazonaws.com/reports/stiffness-21-June-24/resources/jointStiffness9842/index.html

The extra plots where stiffness is limited by distance seem to show that we have to go quite far until the rupture is above the 0.75 threshold.

Reading your comments and talking to Chris here shows to me that I still don't quite understand the stiffness values well enough to properly interpret this. I hope we can clear all that up when we meet next week!

kevinmilner:
@voj oh cool, I better understand what you were getting at with the distances yet. Maybe some sort of intermediate option to balance the near and far field would be to specify the threshold as a function of distance?

voj:
@kevinmilner yes, I was thinking of something like that. I'll whip up something and see how it looks.

voj:
merged with #217

PR #128, opened 2024-10-03
Module archive I/O abstraction

kevinmilner:
When the ModuleArchive (`org.opensha.commons.util.modules.ModuleArchive<E>`) interface was originally created, it was designed to store all module data using the Java ZipFile and ZipOutputStream implementations.

Those implementation have some limitations, notably:

- No ability to zip to and read from memory (useful for sending data over the network in distributed calculations rather than using a slow filesystem)
- Can't copy a zipped entry from one archive to another without decompressing and recompressing

Because of this, various codes use the Apache commons-io zip implementation instead to write large data files (e.g., logic tree hazard maps) and bypass the modules framework.

This PR abstracts the I/O methods of `ModuleArchive` to instead use new `ArchiveInput` and `ArchiveOutput` interfaces. Implementations are also provided using the regular Java classes, Apache commons-IO, and Java NIO. Multiple Apache implementation are provided including in-memory and parallel zipping.

The new `ArchiveOutput.ParallelZipFileOutput` does all of the zipping in memory and in parallel, which can be very fast if you are more zip-bound than write-bound and have lots of available memory. `ArchiveOutput.AsynchronousZipFileOutput` does zipping in the current thread but handles writes asynchronously without too much additional memory overhead.

ModuleArchive's can also be read or written directly to directory structures (unzipped), or to tar files. 

@voj: if you have any custom `ArchivableModule` implementations that implement the `writeToArchive` or `initFromArchive` methods (rather than deferring to one of the default implementations), you will need to modify them once this is merged.


kevinmilner:
Note: developers may have to do a "Gradle -> Refresh Gradle Project" in eclipse to fetch the updated dependencies.

PR #135, opened 2024-10-16
fix issue #134

### LINKED ISSUE #134: Tests Fail
Two tests are failing. It turns out that these two tests call `testModuleSerialization()` twice which opens a zip file as a `ModuleArchive` at the end of the first call and then tries to overwrite it in the second call while the first `ModuleArchive` still has the file open.

This happens on Windows.

```
> Task :opensha:test

org.opensha.sha.earthquake.faultSysSolution.FaultSystemTestSuite > org.opensha.sha.earthquake.faultSysSolution.modules.StandardFaultSysModulesTest.testInvTargetMFDs FAILED
    java.nio.file.FileSystemException at StandardFaultSysModulesTest.java:371

org.opensha.sha.earthquake.faultSysSolution.FaultSystemTestSuite > org.opensha.sha.earthquake.faultSysSolution.modules.StandardFaultSysModulesTest.testRegionsOfInterest FAILED
    java.nio.file.FileSystemException at StandardFaultSysModulesTest.java:371

3654 tests completed, 2 failed
```

voj:
Fixes #134 by closing the `ModuleArchive`'s `'input` explicitly.

@kevinmilner do you think maybe `ModuleArchive` should be `Closeable` instead?

### DIFF
--- a/src/test/java/org/opensha/sha/earthquake/faultSysSolution/modules/StandardFaultSysModulesTest.java
+++ b/src/test/java/org/opensha/sha/earthquake/faultSysSolution/modules/StandardFaultSysModulesTest.java
@@ -444,6 +444,7 @@ public class StandardFaultSysModulesTest {
 		withoutZip.close();
 		withZip.close();
 		rewrittenZip.close();
+		loaded.getInput().close();
 		
 		assertTrue("No unique files found for "+module.getName(), numTests > 0);
 	}


kevinmilner (approved):

kevinmilner:
Sorry for the slow merge on this one! As for `ModuleArchive` implementing `Closeable`, I can see the argument either way. On one hand it would be convenient to quickly close the input, but it can also be constructed in memory without an attached input so that might be confusing. Also it would continue to be usable after closing, except for any unloaded available modules.

PR #140, opened 2025-01-29
Add GetFile to MeanUCERF3

abhatthal:
This PR adds the [GetFile](https://strike.scec.org/scecpedia/GetFile) framework to MeanUCERF3.
GetFile is used to retrieve models from a CARC Globus file share and automatically download new versions as they become available.

New features in this PR:
- GetFile automated updates for MeanUCERF3, UCERF3_CompoundSol_ERF, and UCERF3EpistemicListERF
- CalcProgressBar shows download progress
- Server Status badge now includes CARC Globus availability

Future plans for GetFile include:
- GetFile self-updating JAR
- Allow users to edit client metadata to prompt "Update Now", "Later", "Skip this Version" on per file basis
- Model version snapshot management integrated into UI


src/main/java/org/opensha/commons/geo/Location.java
kevinmilner, line 1: Maybe revert this comment?

Re: your question, that might have unintended consequences. Best to keep it the way it is (I use areSimilar when I cant to know if they're similar, and equals when I want to know if they're exactly equal)

kevinmilner:
The automated builds are failing with this error:

> bad class file: /home/runner/work/opensha/opensha/lib/getfile-all.jar(/org/scec/getfile/GetFile.class)
>    class file has wrong version 65.0, should be 55.0

The tests run with Java 11, which is the lowest version of java we currently support. This error means that you compiled GetFile with a newer version, and didn't tell it to maintain compatibility with Java 11. Can you rebuild GetFile? See the top of OpenSHA's build-common.gradle to see how to set that.

kevinmilner:
I tested out the hazard curve application, both before and after clearing out my old .opensha directory. The downloads were seamless and quick in both cases, well done! MeanUCERF3 worked well, but there seems to be an issue with the Single Branch and Epistemic List versions. After the download finishes, it is supposed to refresh the parameter list in the GUI so that you can choose logic tree levels. Here's what it looks like when you open single branch ERF on the current master branch:

image

Note all of the parameters for branches, starting with Fault Model at the top. That doesn't happen on your fork:

image

Same issue with the epistemic list ERF. Main branch:

image

And yours:

image

Can you look into that?

abhatthal:
"UCERF3 Epistemic List" and "UCERF3 Single Branch ERF" now show retrieved parameters after download in both HazardCurveApplication and HazardSpectrumApplication.

kevinmilner:
Hi Akash, a few comments so far:

1. Can you make ParameterList fire a change list on the replace method as well?

2. MeanUCERF3 seems to be a lot slower. With default settings, instantiating and calling updateForecast() took ~5s before and takes ~24s now. I don't think it's because of download speed, I think what's happening is:

Original flow with default parameter settings:

* check to see if we can do a shortcut to download the pre-cached file directly
* we can, download `cached_FM3_1_dep100.0_depMean_rakeMean.zip`
* load that file, done

Flow in your fork with default parameter settings:

* downloads both `cached_FM3_1_dep100.0_depMean_rakeMean.zip` and `FM3_1_mean_ucerf3_sol.zip`
* doesn't bother using the already cached `cached_FM3_1_dep100.0_depMean_rakeMean.zip` file, instead rebuilds it from scratch in memory from `FM3_1_mean_ucerf3_sol.zip`
* write out that in-memory `cached_FM3_1_dep100.0_depMean_rakeMean.zip` instance, blowing away the version that was just downloaded

So it's skipping the shortcut and also blowing away the version it just downloaded. This happens both on first invocation (where it needs to download) and on subsequent invocations (where it doesn't).

Here is my simple test code to make testing (and benchmarking) this easier:

```java
import java.util.concurrent.TimeUnit;

import com.google.common.base.Stopwatch;

import scratch.UCERF3.erf.mean.MeanUCERF3;

public class MeanU3DownloadAndUpdateBenchmark {

	public static void main(String[] args) {
		Stopwatch watch = Stopwatch.createStarted();
		MeanUCERF3 u3 = new MeanUCERF3();
		
		u3.updateForecast();
		
		watch.stop();
		System.out.println("Took "+watch.elapsed(TimeUnit.SECONDS)+"s");
	}

}
```

You can write it out to a file named `MeanU3DownloadAndUpdateBenchmark.java` and execute it as follows:

```
java -cp /path/to/opensha-all.jar MeanU3DownloadAndUpdateBenchmark.java
```

3. When running that test, it never exits on your fork. That usually happens when some sort of non-daemon thread is started and never stopped. I see that there are multiple thread pools running at the end of execution causing it to never exit, likely created here in the GetFile constructor:

```java
		this.updateFileExec = Executors.newSingleThreadExecutor();
		this.updateAllExec = Executors.newSingleThreadExecutor();
		Runtime.getRuntime().addShutdownHook(new Thread(() -> {
			updateFileExec.shutdown();
			updateAllExec.shutdown();
			meta.writeClientMetaState();
		}));
```

It seems like you're trying to deal with this via the shutdown hook. I'm not familiar with that, but it doesn't seem to be working. Can you look into it?

If possible, I suggest using the static CompletableFuture methods to run things on the fork-join pool instead of custom thread pools (ExecutorService instances). ExecutorService is great if you need to run with a fixed number of threads and you can clean up after (call shutdown()), but otherwise it's not a great tool. If you absolutely must use them, then you can make them daemon threads but that's a bit kludgy.

abhatthal:
1. I updated `setValue` to fire a change event. The `replaceParameter` method invokes `removeParameter` and then `addParameter`, both of which already fire change events.

2. Fixed, now see ~4 seconds to update forecast.

3. Fixed application hanging by correctly disposing CalcProgressBar and using the default Fork-Join pool for all CompletableFutures.

src/main/java/org/opensha/commons/param/ParameterList.java
```diff
@@ -257,56 +260,51 @@ public <T> Parameter<T> getByIndex(Class<T> type, int index) {
...
+		name = getParameterName(name);
 		int index = getIndexOf(name);
-		if( index !=-1 ) {
+		if (index != -1) {
 			Parameter param = (Parameter)params.get(index);
 			param.setValue(value);
-		}
-		else{
-
+			fireChangeEvent(new ChangeEvent(this));
```
kevinmilner, line 271: this method is for setting a parameter value, not changing the parameters contained in the list. please remove the `fireChangeEvent(...)` call.

src/main/java/org/opensha/commons/param/ParameterList.java
```diff
@@ -674,5 +656,43 @@ public static boolean setParamsInListFromXML(ParameterList paramList, Element pa
 
 		return !failure;
 	}
+	
+	
+	/**
+	 * Add a listener to notify about ParameterList events (i.e. add/remove)
```
kevinmilner, line 662: Maybe update this comment to be more specific that listeners will be notified when the contents of the parameter list are changed (e.g., parameters added or removed), but will not be notified when parameter values themselves are changed.

kevinmilner (changes requested):
Looks good. You're right that replaceParameter didn't actually need to explicitly fire an event (it calls remove and add, which do), sorry for my erroneous suggestion.

Only change before merging is that setValue(String, Object) should _not_ fire a change event, those are reserved for the parameters contained in the list changing, not their values changing.

src/main/java/org/opensha/commons/param/ParameterList.java
abhatthal: * Removed the `fireChangeEvent` in `ParameterList.setValue`
```java
	/** Set's a new value to a Parameter in the list if it exists, else throws exception. */
	public void setValue(String name, Object value) throws ParameterException, ConstraintException {
		String S = C + ": setValue(): ";
		if(D) System.out.println(S + "Starting");

		name = getParameterName(name);
		int index = getIndexOf(name);
		if (index != -1) {
			Parameter param = (Parameter)params.get(index);
			param.setValue(value);
		} else {
			throw new ParameterException(S + "No parameter exists named " + name);
		}
	}
```


* Updated javadoc for `ParameterList.fireChangeEvent` to clarify listeners should only be notified about changes to the ParameterList itself.

```java
	/**
	 * Signal to listeners that a change has occured to the ParameterList.
	 * Listeners will be notified when the contents of the parameter list are
	 *  changed (e.g., parameters added or removed), but will not be notified
	 *  when parameter values themselves are changed.
	 * @param event Event that occured (i.e. add/remove parameter)
	 */
	private void fireChangeEvent(ChangeEvent event) {
		if (D) System.out.println("ParameterList.fireChangeEvent()");
		ChangeListener[] listeners;
		synchronized (this) {
			if (changeListeners == null) return;
			listeners = new ChangeListener[changeListeners.size()];
			changeListeners.toArray(listeners);
		}
		for (ChangeListener listener : listeners) {
			listener.stateChanged(event);
		}
	}
```

kevinmilner (approved):

PR #142, opened 2025-02-19
Add NSHM23 Western US Branch Averaged ERF

abhatthal:
Create class loader for USGS 2023 NSHM ERF, Western U.S., Branch Averaged ERF.

This PR introduces two new classes
* NSHM23_Downloader : extends GetFile to specifically download model files for NSHM 2023 from a defined path on the CARC server
* NSHM23_WUS_BranchAveragedERF : extends BaseFaultSystemSolutionERF to load the ERF downloaded by NSHM23_Downloader


abhatthal:
I can see the ERF available in the HazardCurveApplication. Do I need to set any parameters? Since it's a branch averaged solution, I wouldn't think I need to set any. It still uses the default parameters, i.e. solution input file. I'm not sure if these are relevant.
Screenshot 2025-02-18 at 8 42 09 PM

src/main/java/org/opensha/sha/earthquake/rupForecastImpl/nshm23/erf/NSHM23_WUS_BranchAveragedERF.java
```diff
@@ -0,0 +1,83 @@
...
+import org.opensha.sha.earthquake.rupForecastImpl.nshm23.util.NSHM23_Downloader;
+
+/**
+ * USGS 2023 NSHM ERF, Western U.S., Branch Averaged ERF
+ */
+public class NSHM23_WUS_BranchAveragedERF extends BaseFaultSystemSolutionERF {
+
+	private static final long serialVersionUID = 277613161331416141L;
+	private static final String MODEL = "WUS_branch_averaged_gridded_simplified";
+	public static final String NAME = "NSHM23 WUS Branch Avg ERF";
```
kevinmilner, line 20: I realized that we should specify that this ERF doesn't include the Cascadia subduction zone. Please rename to:

"NSHM23-WUS (excl. Cascadia) Branch Avg ERF

kevinmilner (changes requested):
Looks good! I like that the downloader is general for reuse with models for other NSHM23 regions as well.

Besides the minor rename, can you also set it to hide the file parameter? Since this is automatically downloaded, we don't want the "Solution Input File" -> "Browse" parameter/button to show up. You can see how to do this by looking at MeanUCERF3.

src/main/java/org/opensha/sha/earthquake/rupForecastImpl/nshm23/erf/NSHM23_WUS_BranchAveragedERF.java
abhatthal: SGTM, done!

kevinmilner (approved):
Looks good, thanks!

PR #144, opened 2025-03-10
Thread refactoring

abhatthal:
This PR resolves bug #109.

### Changes
- Remove Thread.stop() invocations inside
  - HazardCurveApplication
  - HazardSpectrumApplication
- Check for interrupts both inside the HazardSpectrumCalculator and before/after updating forecast
- Improved error messages for missing ERFs
- Added GCIM calculations into HazardCurveApplication and removed the GCIM Enabled HazardCurveApplication.

### Considerations
- Is DefaultExceptionHandler is needed anymore or if it should be deprecated/removed.
- Other non-distributed applications still use Thread.stop(), should these be deleted?


src/main/java/org/opensha/sha/earthquake/faultSysSolution/erf/BaseFaultSystemSolutionERF.java
```diff
@@ -313,14 +313,21 @@ protected void updateHookBeforeFaultSourceBuild() {
...
 		if (D) System.out.println("Updating forecast");
 		long runTime = System.currentTimeMillis();
 		
+		if (faultSysSolution == null) {
+			if (D) System.out.println("Failed to update forecast, faultSysSolution == null.");
+			throw new NullPointerException(
+					"Failed to update forecast. Fault system solution is unavailable. "
+					+ "Ensure the ERF is provided prior to computation.");
+		}
+
```
kevinmilner, line 327: shouldn't this be after the `if (fileParamChanged) { readFAultSystemSolutionFromFile(); }` code? As currently written, I don't think the file based version of the code could ever work (you can test by selecting "Fault System Solution ERF" and browsing for a solution file).

src/main/java/org/opensha/sha/calc/SpectrumCalculator.java
```diff
@@ -204,7 +205,10 @@ public DiscretizedFunc getIML_SpectrumCurve(DiscretizedFunc spectrumFunction,
 
 		// loop over sources
 		for (sourceIndex = 0; sourceIndex < numSources; sourceIndex++) {
-
+			// quit if user cancelled the calculation
+			if (HazardCurveApplication.isCancelled())
+				return new ArbitrarilyDiscretizedFunc();
+
```
kevinmilner, line 211: I don't love this static approach here; I would prefer that the application tells the calculator that a cancellation has been requested. Here's what I proposed in my comment on #109:

> The simplest might be to add an AtomicBoolean to HazardCurveCalculator that represents a canceled state with a setter method, and to check if it has been set inside the source loop (probably not the inner rupture loop) of computeHazardCurve and exit if so.

More specifically, add a private `AtomicBoolean` instance to HazardCurveCalculatator that defaults to false and setter method to indicate cancellation (e.g., `setCancelled()`). Then, if you detect cancellation in the loop, reset it to to false and return null (don't return an empty function).

src/main/java/org/opensha/sha/gui/HazardCurveApplication.java
```diff
@@ -162,7 +157,7 @@ public class HazardCurveApplication extends JFrame implements
 	 */
 	private final static String C = "HazardCurveServerModeApplication";
 	// for debug purpose
-	protected final static boolean D = false;
+	protected final static boolean D = true;
```
kevinmilner, line 160: will want to revert this before release (OK to leave true for now)

src/main/java/org/opensha/sha/gui/HazardCurveApplication.java
```diff
@@ -2267,46 +2244,71 @@ public void print() {
...
+	 * Currently this is a volatile boolean, but alternative signallers
+	 * may be considered if we need multiple concurrent writers.
+	 * @return If the signal to cancel the Hazard Curve calculation is sent
+	 */
+	public static boolean isCancelled() {
+		if (D && cancelled) {
+			System.out.println("Caught cancellation signal");
+		}
+		return cancelled;
+	}
```
kevinmilner, line 2259: see prior comments about setting cancel signal instead in the calculator itself

kevinmilner (changes requested):
See comments for suggestions.

I also only see cancellation implemented for the `SpectrumCalculator`; I expected to see `HazardCurveCalculator` updated to detect cancellations (ideally with its own internal AtomicBoolean instance and not the static approach).

I didn't mention this before, but you should also add cancellation support to `DisaggregationCalculator` (checking for cancellation at the source level) as that can be an expensive calculation as well.

src/main/java/org/opensha/commons/data/WeightedValue.java
```diff
@@ -9,4 +11,12 @@ public WeightedValue(E value, double weight) {
...
\ No newline at end of file
+	
+	@Override
+	public String toString() {
+		if (value instanceof IntensityMeasureRelationship) {
+			return ((IntensityMeasureRelationship) value).getName();
+		}
+		return value.getClass().getName();
+	}
+}
```
kevinmilner, line 22: Oof, I don't like this hack to make `WeightedValue` know about a specific implementation (`IntensityMeasureRelationship`). How about you check if `value implements Named` and then return `((Named)value).getName()`?

Actually probably best to do this hierarchy:

```java
if (value instanceof ShortNamed)
    return ((ShortNamed)value).getShortName();
else if (value instanceof Named)
    return ((Named)value).getName();
else
    return value.toString(); // better than just the class name, as implementation may already give good toStrings
```

This will accomplish what you're after because `IntensityMeaureRelationship` implements `ShortNamed` (which in turn extends `Named`)

Issue #147, opened 2025-03-28
Unable to set TreeBranchWeightsParam in WG02_FortranWrappedERF EpistemicList.


abhatthal:
Following the OpenSHA tutorial (https://opensha.org/Tutorials) to replicate Figure 5 in https://opensha.org/resources/Field_2005_SRL.pdf.
Resulted in a StackOverFlow error due to the TreeBranchWeightsParam.setValue calling itself recursively instead of setting the value.

Validate that after setting the value correctly, the generated plot is as expected.

<img width="1327" alt="Image" src="https://github.com/user-attachments/assets/69c4981a-90b2-47d9-b6d1-ade4c2a04b66" />

abhatthal:
The inability to set parameters was resolved in #148. However, the Fortran backend for the WG02_FortranWrappedERF is unable to parse changes, and as such generated plots do not differ. There are no current plans to fix this, as it requires reviewing legacy Fortran code. The fix for ParameterListParameters can be generally applicable to any ERF that leverages them, so the fix will still be merged in.

PR #148, opened 2025-03-28
Tree branch weights stack overflow

abhatthal:
Fix for #147 

Fixes StackOverflowError recursion, ensure TreeBranchWeightParameters are set correctly, and clean up old code.

### DIFF
--- a/src/main/java/org/opensha/commons/param/impl/ParameterListParameter.java
+++ b/src/main/java/org/opensha/commons/param/impl/ParameterListParameter.java
@@ -21,7 +21,6 @@ import org.opensha.commons.param.editor.impl.ParameterListParameterEditor;
 
 public class ParameterListParameter extends AbstractParameter<ParameterList> {
 
-
 	/**
 	 * 
 	 */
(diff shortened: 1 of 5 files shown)


abhatthal:
I'm unable to replicate Figure 5 by invoking super.setValue.
fig-5-plots-dont-match

abhatthal:
Fixes ParameterListParameter setValue, but Fortran backend doesn't parse values correctly. Merging relevant OpenSHA changes in. More details in #147.

src/main/java/org/opensha/commons/param/impl/ParameterListParameter.java
```diff
@@ -51,118 +50,51 @@ public ParameterListParameter(String name) {
...
-		param.editable = true;
+		try {
+			if (value == null)
+				param = new ParameterListParameter(name);
+			else
+				param = new ParameterListParameter(name,(ParameterList)value);
+			param.editable = true;
+		} catch(Exception e) {
+			e.printStackTrace();
+			param.editable = false;
```
kevinmilner, line 88: What's the motivation for this?

abhatthal: In both cases, param should be assigned to a new ParameterListParameter, so there's no case where it would be null unless construction failed (i.e. ConstraintException thrown in AbstractParameter).

Instead of silently returning null, this catches and prints the exception and reflects the fact that the param can't be edited as it's null.

kevinmilner: Did you encounter it returning null?

The way I read the old code, the `if( param == null ) return null;` could never hit because `param` was assigned by either clause of the if/else, and there was no try/catch.

Your revised code I think is trying to return null if there's an exception, but it feels better to me for the exception to just be propagated to the caller as it was before. Also, I think your try/catch would introduce an NPE because you call `param.editable = false;` in a try/catch that can only be triggered if there's an instantiation error on the  `ParameterListParameter` constructor. In other words, the only way to get into the catch block results in `param` being null, and then you assign a variable inside `param` which would throw an NPE.

Unless I'm missing something, I would just remove the `if( param == null ) return null;` clause that can't be hit in the original method and otherwise keep as is.

PR #181, opened 2025-07-08
Fix transient Server Status failure

abhatthal:
Resolved a transient TimeOutException in OpenSHA’s GetFile operational tests.

Sometimes we’d get a TimeOutException, as the GitHub Actions server is unable to download a 67MB file for the MeanUCERF3 metadata from CARC in under 20 seconds. We’re still able to establish a connection with the ERF data files for UCERF3 and NSHM23.

This is only a transient bug (see https://github.com/opensha/opensha/actions/runs/16114950142), possibly due to an unusually large load on one of or both of the GitHub Actions server and CARC.
I’ve since changed this to only verify we can establish a connection to both the UCERF3 and NSHM23 meta files
and created a manual (non-operational) test to verify we’re able to download all files. This is useful in determining if all files can be downloaded and we’re not impacted by file system corruption.


kevinmilner (commented):
Looks good. The update all test could potentially be improved by removing the hardcoded file names and grabbing them from the metadata file instead.

abhatthal:
Accidentally copied tests branch into opensha instead of in fork. Closing PR and opening new one.

PR #187, opened 2025-11-03
[186] fix legend cut-off

### LINKED ISSUE #186: Graph legend cut off in some reports
As originally reported in https://github.com/GNS-Science/nzshm-opensha/issues/389, some report charts have their legend cut off:

Image

voj:
closes #186 

For `SectBySectDetailsPlots`, the legend was anchored at 0.975 (i.e. 97.5%) of the X axis whenever Latitude was mapped to X, and 0.025 whenever Longitude was mapped to X. This change always anchors the legend at 0.025, i.e. the legend begins at the left-hand side of the chart rather than the right-hand side of the chart.

Going through the code, I do not believe that my change breaks anything. It is not obvious why the legend's origin was set to the far right of the chart. It's possible that this was done because the X axis is inverted when Latitude is mapped to X, and the hope was that the anchor would then be equivalent to 0.025. Another theory is that the legend was supposed to run vertically on the righ-hand side of the chart, but I cannot see any code to rotate the legend.

Before and after:

sect_along_strike

sect_along_strike

### DIFF
--- a/src/main/java/org/opensha/sha/earthquake/faultSysSolution/reports/plots/SectBySectDetailPlots.java
+++ b/src/main/java/org/opensha/sha/earthquake/faultSysSolution/reports/plots/SectBySectDetailPlots.java
@@ -2331,7 +2331,7 @@ public class SectBySectDetailPlots extends AbstractRupSetPlot {
 			maxMag = Math.max(maxMag, rupSet.getMaxMagForSection(sect.getSectionId()));
 		}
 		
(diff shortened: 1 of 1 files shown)


kevinmilner (approved):
Good catch! I think there must have been another bug at some point that led to that not showing up for me in NSHM23 plots where latitudeX is true, so I never caught it. But I can reproduce it now, so something else must have changed.

Sorry for the delay merging this in, just getting back up and running following the government shutdown.

PR #190, opened 2025-11-20
IM Event Set Calculator: Fix GUI (2/4)

abhatthal:
This is 2/4 PRs for the IM Event Set Calculator.
This is the largest PR for this feature. Future PRs will be much smaller.

**Feature Development Plan:**
https://docs.google.com/document/d/1W3IBBzcrgg1A2ypIkE11r3_vKqojrWG9pjvgRjTInZo/edit?usp=sharing

**Summary of PR Changes:**
* Fix critical bugs allowing GUI application to be useable
* Allow Site Data Types list to be generated from union of IMRs
* Improve UI flow and add warning messages for removing IMRs/IMTs.
* Remove load/save GUI calc buttons and corresponding logic
* Create progress bar in bottom panel
* Change default behavior to enable all site data providers
* Create unit tests for ParameterList intersection and union methods
* Add EULA dialog and set to use OpenSHA build version instead of v3

### DIFF
--- a/src/main/java/org/opensha/sha/calc/IM_EventSet/v03/gui/SitesPanel.java
+++ b/src/main/java/org/opensha/sha/calc/IM_EventSet/v03/gui/SitesPanel.java
@@ -25,6 +25,7 @@ import javax.swing.event.ListSelectionListener;
 
 import org.opensha.commons.data.siteData.OrderedSiteDataProviderList;
 import org.opensha.commons.data.siteData.SiteData;
+import org.opensha.commons.data.siteData.SiteDataValue;
 import org.opensha.commons.data.siteData.gui.beans.OrderedSiteDataGUIBean;
 import org.opensha.commons.data.siteData.impl.WillsMap2000;
 import org.opensha.commons.exceptions.InvalidRangeException;
@@ -33,22 +34,24 @@ import org.opensha.commons.param.Parameter;
 import org.opensha.commons.param.ParameterList;
 import org.opensha.sha.imr.IntensityMeasureRelationship;
 import org.opensha.sha.imr.ScalarIMR;
+import org.opensha.sha.util.SiteTranslator;
 
 public class SitesPanel extends JPanel implements ListSelectionListener, ActionListener {
-	
-	protected JList sitesList;
-	protected JList siteDataList;
-	
-	protected JButton addSiteButton = new JButton("Add Site");
-	protected JButton removeSiteButton = new JButton("Remove Site(s)");
-	protected JButton editSiteButton = new JButton("Edit Site");
-    protected JButton importSitesButton = new JButton("Import Sites");
-    protected JButton exportSitesButton = new JButton("Export Sites");
+
+	private final JList sitesList;
+	private final JList siteDataList;
+
+	private final JButton addSiteButton = new JButton("Add Site");
+	private final JButton removeSiteButton = new JButton("Remove Site(s)");
+	private final JButton editSiteButton = new JButton("Edit Site");
+    private final JButton importSitesButton = new JButton("Import Sites");
+    private final JButton exportSitesButton = new JButton("Export Sites");
+    private final JButton setFromWebButton = new JButton("Fetch Site Data");
 
     // List of locations for each site
-	private ArrayList<Location> locs;
+	private final ArrayList<Location> locs;
     // List of all site data parameters for each site
-    private ArrayList<ParameterList> siteDataParams;
+    private final ArrayList<ParameterList> siteDataParams;
     // Default site data parameters to use for new sites
     private ParameterList defaultSiteDataParams;
     // Set sites with selected site data providers
@@ -67,7 +70,7 @@ public class SitesPanel extends JPanel implements ListSelectionListener, ActionL
         siteDataParams = new ArrayList<>();
         defaultSiteDataParams = new ParameterList();
 
-		sitesList = new JList();
+		sitesList = new JList<>();
 		sitesList.setSelectionMode(ListSelectionModel.MULTIPLE_INTERVAL_SELECTION);
 		sitesList.setSelectedIndex(0);
 		sitesList.addListSelectionListener(this);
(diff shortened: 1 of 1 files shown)


kevinmilner:
There appears to be a bug where the site data values aren't always actually being used.

I set up a test calculation with 3 sites, where I first disabled all site data providers.

Site 1: 34.0, -118.0
Vs30=760
Z2.5=null
Z1.0=null

Site 2: 34.0, -118.00001 (essentially the same)
Vs30=260
Z2.5=null
Z1.0=null

Site 3: 34.0, -118.00002 (essentially the same)
Vs30=260
Z2.5=10
Z1.0=3000

I used NGA West2 (No Idriss) as the IMR, and NSHM23 with background seismicity excluded. The IMTs are PGA, 1s SA, and 5s SA. Looking at the event file outputs, the first 2 sites with the Vs30 change are indeed different as expected, but the basin depth change from site 2 to 3 is not having any effect:

```
kevin@steel:/tmp/im_test$ head -n5 NGAWest_2014_NoIdr_*
==> NGAWest_2014_NoIdr_PGA.txt <==
68 0 -5.4047 0.6167 0.3325 -4.8335 0.6067 0.3307 -4.8335 0.6067 0.3307
69 0 -5.2005 0.6151 0.3295 -4.6313 0.6046 0.3273 -4.6313 0.6046 0.3273
70 0 -5.0575 0.6139 0.3273 -4.4899 0.603 0.3248 -4.4899 0.603 0.3248
71 0 -4.9511 0.613 0.3256 -4.3849 0.6017 0.3228 -4.3849 0.6017 0.3228
72 0 -4.8649 0.6126 0.3247 -4.2999 0.6009 0.3217 -4.2999 0.6009 0.3217

==> NGAWest_2014_NoIdr_SA_1.0.txt <==
68 0 -4.8257 0.7248 0.3433 -3.8319 0.7211 0.3423 -3.8319 0.7211 0.3423
69 0 -4.6199 0.7235 0.3405 -3.6273 0.7194 0.3392 -3.6273 0.7194 0.3392
70 0 -4.4761 0.7225 0.3384 -3.4847 0.7182 0.3369 -3.4847 0.7182 0.3369
71 0 -4.3694 0.7217 0.3368 -3.3789 0.7172 0.3351 -3.3789 0.7172 0.3351
72 0 -4.2831 0.7213 0.3359 -3.2934 0.7165 0.3341 -3.2934 0.7165 0.3341

==> NGAWest_2014_NoIdr_SA_5.0.txt <==
68 0 -6.5861 0.7159 0.3733 -5.6101 0.7159 0.3733 -5.6101 0.7159 0.3733
69 0 -6.2767 0.7145 0.3706 -5.3007 0.7145 0.3706 -5.3007 0.7145 0.3706
70 0 -6.0539 0.7135 0.3687 -5.0779 0.7135 0.3687 -5.078 0.7135 0.3687
71 0 -5.885 0.7128 0.3672 -4.9091 0.7128 0.3672 -4.9091 0.7128 0.3672
72 0 -5.7517 0.7124 0.3664 -4.7758 0.7124 0.3664 -4.7758 0.7124 0.3664
```

Notice how values for sites 2 and 3 are always identical; that should not be the case, especially for 5s SA. For comparison, here is the output of the hazard curve application for the same IMR, ERF, and site for 5s SA. Red is the curve with null basin depth params, and blue is the curve with Z2.5=10km and Z1.0=3000m.

image

kevinmilner:
Great progress, we're getting close! A few notes follow in this and subsequent comments.

The current output list includes many "-1" values for skipped ruptures for individual sites. Was that part of the prior command line application behavior? My instinct is to just completely skip writing data for any source/rupture combinations that are skipped for all selected sites, and then to use NaN rather than -1 for skipped values (-1 is a valid IML, so that being the "no-data" value can cause confusion). Make sure to also replace the "-1" with "NaN" for the case where an IMR doesn't have the std. dev. type param (e.g., WrappedASK_2014) and the 3rd output column is currently set to -1

 If -1 was there before I could maybe be convinced to keep it the same for that output writer, but I think NaN is clearer

kevinmilner:
Definitely a bug and probably related to the above comment: it appears that you didn't clone the site parameter objects for each site instance and thus they're shared. Each site needs it's own cloned version of each site parameter so that their values can be adjusted independently of other sites, and so that changing an IMR's internal site parameter settings doesn't propagate that setting back to modify any of the individual site's parameters. I notice that, e.g., if I manually change the Z2.5 value on one site in the edit site GUI, it updates the value on all sites; that shouldn't occur.

Related: if you hit "Edit Site" and change a site data value, but then hit cancel, it still updates the value for that site (and every other site because of the above bug). That edit doesn't show up in the "Site Data Values" list until you click to another site and then click back to force it to refresh.

abhatthal:
> Great progress, we're getting close! A few notes follow in this and subsequent comments.
> 
> The current output list includes many "-1" values for skipped ruptures for individual sites. Was that part of the prior command line application behavior? My instinct is to just completely skip writing data for any source/rupture combinations that are skipped for all selected sites, and then to use NaN rather than -1 for skipped values (-1 is a valid IML, so that being the "no-data" value can cause confusion). Make sure to also replace the "-1" with "NaN" for the case where an IMR doesn't have the std. dev. type param (e.g., WrappedASK_2014) and the 3rd output column is currently set to -1
> 
> If -1 was there before I could maybe be convinced to keep it the same for that output writer, but I think NaN is clearer

Done. Now writes "NaN" instead of "-1" and the line is skipped altogether if all sites for the rupture have only NaN values

abhatthal:
> Export sites view: I would auto-populate it with all site data columns by default. I suppose also make the import default match the export (with all site data columns enabled) so that one can quickly re-import a previously-exported site list.
> 
> I would also use this format for the `site_metadata.txt` file because it's more machine readable and will allow users to quickly reload their site list from an earlier calculation. For that file, always include all site parameters.
> 
> I notice that if you hit OK in this view without specifying an output file, you get an "I/O error reading file" message that seems to be left over from the import view.

We'll use the auto-generated machine-readable CSV for import/export we discussed in future version.
Fixed the output error message for invalid/missing output file.
Both the import and export panels are pre-populated with values Vs30,Z1.0,Z2.5

PR #191, opened 2025-12-17
Update SimpleVulnerability DEM

annehulsey:
Updates the SimpleVulnerability class to create the damage exceedance matrix needed for the LossEstimationApplication.

Reads Keith's vulnerability files and assumes a lognormal distribution for the mean and cov values.

Creates a new VulnerabilityLocalAccessor to use during troubleshooting. (The current state uses the servlet. This can be switched by commenting either line 294 or 295 in LossEstimationApplication.)

### DIFF
--- a/src/main/java/org/opensha/sra/vulnerability/models/SimpleVulnerability.java
+++ b/src/main/java/org/opensha/sra/vulnerability/models/SimpleVulnerability.java
@@ -22,7 +22,7 @@ public class SimpleVulnerability extends AbstractVulnerability implements Serial
 	/**
 	 * 
 	 */
-	private static final long serialVersionUID = 1L;
+	private static final long serialVersionUID = 2L;
 	
 	private ArrayList<Double> mdfVals;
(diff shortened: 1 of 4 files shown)


src/main/java/org/opensha/sra/vulnerability/models/SimpleVulnerability.java
kevinmilner: Looks great. Can you bump `private static final long serialVersionUID = 1L;` to `2L`? That flag change will indicate that the class and how it stores its data has changed, and is checked when trying to access the servlet. Once I merge this in and update the servlet, and both the servlet and the main codebase are on the same version (2), things should work well. If you leave it set to 1, but still change the underlying data, weird things will happen when trying to access the servlet.

kevinmilner (changes requested):
Good work! Once you make that minor version change, I'll merge it in and also update things server-side.

kevinmilner (approved):

PR #195, opened 2026-02-10
Remove WG02_Fortran ERF

abhatthal:
Remove entries for WGCEP 2002 Fortran Wrapped ERF


abhatthal:
Kept all the original ERF classes and servlet mappings. Simply commented it out from the ERF_Ref, so it's easy to add back if we should ever want to.

PR #197, opened 2026-02-11
IM_EventSetCalc Restructure

abhatthal:
abhatthal/opensha-fork/feature/im-event-set-calc/03-restructure
--> opensha/opensha/master

- Removes the old v1 and v2 variants of IM Event Set Calc
- Renames classes to be more intuitive
- Improve documentation with class-level Javadocs
- Modify Gradle build tasks to include IM Event CLT and GUI

Further improvements to the CLT are emphasized in the following PR (abhatthal/opensha-fork/feature/im-event-set-calc/04-new-cli)
There are also some GUI changes tacked on in there (i.e. CSV output support), as they should occur after this restructure PR.


abhatthal:
> Looks fine to me. I didn't quite follow the changes to ShakeMap app GUI, but I trust that you verified that it still works?

Yes, it still works. I removed the IM_EventSetCEA_ControlPanel from the ShakeMap app Control Panels list. This utilizes the IM_EventSet.v01.IM_EventSetScenarioForCEA. There is no equivalent in later versions of the application. I believe we discussed this was for the Puente Hill Scenario, which we no longer have a need for.

PR #198, opened 2026-02-12
Recently added SourceUtil class belongs in the filters package

kevinmilner:
The  `HazardCurveCalculator.canSkipSource/Rupture(...)` methods were moved to a new `org.opensha.sha.util.SourceUtil` class in a recent PR, which is a good improvement as they're of general use.

This PR does some additional refactoring to put things in more consistent packages. `SourceUtil` is now `org.opensha.sha.calc.sourceFilters.SourceFilterUtils`.

It also moves the source filters from `org.opensha.sha.calc.params.filters` to `org.opensha.sha.calc.sourceFilters` to reflect that there are more than just parameter classes. The parameters themselves are now in a new `org.opensha.sha.calc.sourceFilters.params` subpackage, including the old `MaxDistanceParam`, `MinMagnitudeParam`, etc.

### DIFF
--- a/src/main/java/org/opensha/sha/gui/controls/PEER_TestCaseSelectorControlPanel.java
+++ b/src/main/java/org/opensha/sha/gui/controls/PEER_TestCaseSelectorControlPanel.java
@@ -297,8 +297,9 @@ public class PEER_TestCaseSelectorControlPanel extends ControlPanel {
 		ParameterList siteParams = siteGuiBean.getParameterListEditor().getParameterList();
 
 		// set the distance in control panel
-		SourceFilterManager filterManager = (SourceFilterManager)application.getCalcAdjustableParams()
-				.getParameter(SourceFiltersParam.NAME).getValue();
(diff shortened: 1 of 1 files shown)


abhatthal:
> Good catches, thanks!
> 
> Your proposed solutions sidestep the issue allowing it to fail silently rather than doing what was intended (set the max site-source distance) because `calc.getAdjustableParams().containsParameter(MaxDistanceParam.NAME)` will always return false. I just committed fixes that restore the original functionality.

One last thing, while it is now setting the Fixed Distance Cutoff max distance correctly, it doesn't update the view in the Control Panel if it was initialized before selecting the PEER test case. You have to uncheck and recheck the "Fixed Distance" selection in the Control Panel to see the value update from 200 to 300.

For a good separation of concerns, I don't think we should pass control panels into each other, but maybe we could reinitialize the CalculationSettingsControlPanel from the HazardCurveApplication.selectControlPanel after a user selects a case from the PEER test case control panel? Otherwise we could use an event listener to signal from the PEER control panel to tell the calc settings control panel to refresh itself.

PR #202, opened 2026-02-23
Make GeoJSONFaultSection not final

voj:
`GeoJSONFaultSection` can currently not be mocked for tests as it is `final`. Removing that keyword enables us to write more concise tests.

```
org.mockito.exceptions.base.MockitoException: 
Cannot mock/spy class org.opensha.sha.faultSurface.GeoJSONFaultSection
Mockito cannot mock/spy because :
 - final class
```


voj:
> I had initially made if final because I wanted it to be clear that this class will be loaded as is with the FSS file format and all functionality should be based on its contents rather than extending it.
> 
> I haven't used Mockito before so forgive my ignorance, but would it work if we removed the final keyword but replaced the currently-public constructor (`public GeoJSONFaultSection(FaultSection sect)` with a static `GeoJSONFaultSection.fromSect(FaultSection sect)` method? That would leave it with no public constructors and effectively final, even without the final keyword.

@kevinmilner that should work. I'll try it out.

PR #217, opened 2026-03-10
MultiRupture Generation

voj:
This is a new and tidied-up version of #118. That PR has some of our dicsussion in the history.

This PR introduces experimental code to generate ruptures multi ruptures: ruptures that contain both crustal and subduction sections. The basic approach is to take a rupture set with exisiting crustal and subduction ruptures and then check if combinations of crustal and subduction ruptures are plausible. This PR includes a filter devised by Kevin, and one by Bruce.

See the `readme.md` file for a write-up of the main components.

These changes include some NZ-specific code, mainly a way to distingiush between crustal and subduction fault sections. I think that's OK for now as it's experimental. 

We might want to consider adding such a property to FaultSection in general. In our newer code, we use this class to add properties to sections. https://github.com/GNS-Science/nzshm-opensha/blob/main/src/main/java/nz/cri/gns/NZSHM22/opensha/ruptures/FaultSectionProperties.java The "partition" property can be used to determine if a section is crustal.


voj:
@kevinmilner I've changed subduction detection to use TRTs now 👍

kevinmilner (approved):

PR #227, opened 2026-05-06
Chore: updated gradle wrapper update from 8.4 to 9.5.0

kevinmilner:
This updates the gradle wrapper from 8.4 to the latest 9.5.0, including fixing deprecation warnings for both v9 and forthcoming v10 releases.

This fixes some issues using newer java versions, e.g., Java 21 is not supported for running gradle 8.4. OpenSHA remains on Java 17, but Java 21 is often installed and could cause issues with the outdated gradle version and building in eclipse.

Similar PRs will follow for opensha-dev, opensha-cybershake, and opensha-oaf.

CC: @pmpowers-usgs @voj


kevinmilner:
This change should be pretty transparent for developers; you may have to refresh the gradle project, but it otherwise worked out of the box when I tested pulling this branch on another computer. Downstream projects may need to make corresponding changes if their gradle scripts use any deprecated features (not the case for any of our opensha-* downstream projects).

PR #237, opened 2026-07-21
OpenSHA Application Updater Subsystem

abhatthal:
image

Adds a GitHub-release-based auto-updater for distributed OpenSHA GUI apps. On each app's launch, a daemon thread checks the running version against the latest release and, with the user's consent, downloads the new JAR, verifies its SHA-256 against the published digest, launches it detached, and exits the old JVM. It never blocks startup, is a no-op in headless/IDE/dev builds, and swallows its own exceptions. Full design notes in `src/main/java/org/opensha/commons/util/updater/README.md`.

## Features
- Update-available prompt (Swing) with two choices: **Update now** and **Remind me later** (7-day defer). There is a disabled third choice, **Skip this version** (button dormant behind a flag by default; plumbing still reachable).  The ability to skip versions can be easily added by toggling this flag.  Release notes render as Markdown in an HTML view, trimmed to the "Release Notes" section when present.
- Atomic download to a `.part` file, SHA-256 verification with 3 retries, then launch of the new JAR; old JAR kept as fallback.
- Skips re-download when the latest versioned JAR is already on disk.
- On the new version's first startup, prompts once whether to delete older JAR(s); deletion is never automatic.

## Structure
- `org.opensha.commons.util.http` — `GitHubClient` (stateless HTTP transport + GSON → `GitHubRelease`/`GitHubAsset` DTOs).
- `org.opensha.commons.util.updater` — `ApplicationUpdater` (workflow), `UpdatePrompt` / `SwingUpdatePrompt` (UI), `AssetDownloader` / `AssetLauncher` (`@FunctionalInterface` seams, method-referenced in production, faked in tests). `MarkdownUtils` gains `renderMarkdownToHtml` / `sectionFromHeading`.
- 7 GUI app `main`s wired via `checkForUpdatesDefault(appName, appShortName, assetPrefix)`.

## Testing
- Offline suites run by default via `./gradlew test`: `HttpSuite`
- (`GitHubClientTest`, JSON-fixture parsing), `UpdaterSuite`
- (`ApplicationUpdaterTest`, 26 tests — version parsing/fallback, asset selection, SHA-256, the prompt decision flow incl. skip-version, already-installed detection, cleanup logic; `GitHubClient` mocked, download/ launch injected as fakes — no network, no `System.exit`)
- `UtilSuite` (`MarkdownUtilsTest`, 10 tests).
- A live `GitHubClientOperationalTest` hits the real API only via `./gradlew testOperational`.
- The real launch/exit is not unit-tested (no-op launcher injected) and should be validated manually, Windows first.

### Manual test on Apple Silicon macOS Sequoia 15.6.1
- Building a JAR pinned to older build version shows the update prompt
- Skipping / Remind later dismisses the prompt with persistence
- Update prompt is only present if the latest version isn't present in the same directory
- Building a JAR with the latest version does not show the update prompt
- Agreeing to download results in the latest JAR being downloaded as expected
- On the first run of the new JAR, old JARs for the same app are prompted for cleanup


abhatthal:
## Manual test on x86_64 Windows 11
Just tested Hazard Curve and Hazard Spectrum apps on Windows 11 (x86_64) with Oracle JDK 25.
- Two JARs per app, build versions fixed to 26.0.0 and 26.1.1.
- Remind later dismisses the prompt with persistence
- Update prompt is only present if the latest version isn't present in the same directory
- Building a JAR with the latest version does not show the update prompt
- Agreeing to download results in the latest JAR being downloaded as expected
- On the first run of the new JAR, old JARs for the same app are prompted for cleanup

abhatthal:
## Manual test on x86_64 Fedora Linux 44
Just tested Hazard Curve and Hazard Spectrum apps on Fedora 44 GNU/Linux 7.0.14-201.fc44.x86_64 with Oracle JDK 25.
- Passed the same criteria as the Windows 11 and macOS Apple Silicon manual tests. (See above)

kevinmilner:
Nothing stands out upon first look, but I'd like to test it somehow. Can you post test app jars somewhere so that I can try it out?

Also, how does it work when apps are launched within eclipse/intellij (not from a jar). Does it fail silently or do any sort of prompting?

abhatthal:
> Nothing stands out upon first look, but I'd like to test it somehow. Can you post test app jars somewhere so that I can try it out?
> 
> Also, how does it work when apps are launched within eclipse/intellij (not from a jar). Does it fail silently or do any sort of prompting?

Attached are JARs I used for testing across Windows, macOS and Linux.
HazardCurveGUI and HazardSpectrumGUI fixed to build 26.0.0 and 26.1.1, but all actually have the latest code with the auto-update feature. You should be able to replicate my findings with these JARs (See my above comments for manual testing)

https://drive.google.com/drive/folders/1v8T62yWSN20NycWGUBUdYaZlQZZf3_R7?usp=sharing


The ApplicationUpdater::checkForUpdates checks if it isn't running from a JAR and will not attempt to update in that case.

mbarall:
I suggest that you provide a way for the user to completely disable the application updater. Ideally, a user should never see the updater dialog if they have disabled the updater. (So, while a "don't ask again" button in the dialog would be a step forward, I suggest that the user be able to disable the updater without ever seeing the dialog.) Of course, the user should have a way to enable the updater again if they so choose.

Disabling the updater should apply globally across all OpenSHA applications, so the user does not need to disable the updater separately for each one.

mbarall:
Also, can you confirm that the application updater will never be triggered unless the `main` function calls `ApplicationUpdater.checkForUpdatesDefault`? This is important for opensha-oaf, as we have several applications that run from JARs and we would not want the application updater to be triggered.

abhatthal:
> Also, can you confirm that the application updater will never be triggered unless the `main` function calls `ApplicationUpdater.checkForUpdatesDefault`? This is important for opensha-oaf, as we have several applications that run from JARs and we would not want the application updater to be triggered.

Yes, currently the user will only ever be prompted to update under the following conditions:

- They are running one of the GUI applications we distribute as part of the OpenSHA releases
  - AttenuationRelationshipGUI
  - GMTMapApp
  - HazardCurveGUI
  - HazardSpectrumGUI
  - IMEventSetCalculatorGUI
  - ShakeMapGUI
  - SiteDataGUI
- They are running a JAR, not through an IDE
- Not in a headless environment (e.g., a server)
- There is a newer stable OpenSHA release available (not alpha/beta) (happens about 1-2x per year)
- The latest version of this application isn't already installed in the same directory as the running JAR is located
- The user did not previously prompt to defer update reminders (defers by 7 days per application)

abhatthal:
> I suggest that you provide a way for the user to completely disable the application updater. Ideally, a user should never see the updater dialog if they have disabled the updater. (So, while a "don't ask again" button in the dialog would be a step forward, I suggest that the user be able to disable the updater without ever seeing the dialog.) Of course, the user should have a way to enable the updater again if they so choose.
> 
> Disabling the updater should apply globally across all OpenSHA applications, so the user does not need to disable the updater separately for each one.

I've written the logic to skip an update altogether for a specific application. It is currently disabled in the UI, but once enabled the user could skip the update (prompts for each wired Java app (see my previous comment for supported apps)). Given that we have stable updates 1-2x per year, this would mean you only see a prompt 1-2x per year per application. This is much less frequent than handling the "remind later" 7 days deferment.

Do you think this would still be too annoying? We could have something like a `~/.opensha/config.json` with a boolean toggle to `disableUpdatePrompts`?

mbarall:
> > I suggest that you provide a way for the user to completely disable the application updater. Ideally, a user should never see the updater dialog if they have disabled the updater. (So, while a "don't ask again" button in the dialog would be a step forward, I suggest that the user be able to disable the updater without ever seeing the dialog.) Of course, the user should have a way to enable the updater again if they so choose.
> > Disabling the updater should apply globally across all OpenSHA applications, so the user does not need to disable the updater separately for each one.
> 
> I've written the logic to skip an update altogether for a specific application. It is currently disabled in the UI, but once enabled the user could skip the update (prompts for each wired Java app (see my previous comment for supported apps)). Given that we have stable updates 1-2x per year, this would mean you only see a prompt 1-2x per year per application. This is much less frequent than handling the "remind later" 7 days deferment.
> 
> Do you think this would still be too annoying? We could have something like a `~/.opensha/config.json` with a boolean toggle to `disableUpdatePrompts`?

I'm thinking of situations where an application might be used in a classroom or training setting, where you wouldn't want students to ever encounter a prompt to update. Or, a user who is working on a project might want to make sure that no update happens in the middle of the project. Or, some users will just not like auto-updates even if they are infrequent.

abhatthal:
I've added the ability to disable the entire OpenSHA Application Auto-Updater across all applications by setting a boolean `disableUpdatePrompts` inside `$HOME/.opensha/config.json`. An invalid configuration will not break applications and a missing or broken config will default to `disableUpdatePrompts=false`, so the system is enabled by default. This disables all prompts from this subsystem, including any prompts to clean up outdated files.

The user must manually create such a file and instructions to do so have been added to the update prompt.

I've also added the button to skip a release altogether (previously the button was disabled although logic was already there).

I've updated the assets in the v26.8.0-alpha Google Drive with the latest changes, so you can now validate the behavior to disable update prompts.
I've also written a brief utility to clear Java prefs to "Remind later" or "skip", as well as view preferences and toggle the OpenSHAConfig disableUpdatePrompts state. I've added the fatJar to make this accessible.
(e.g., `java -cp opensha-all.jar  org.opensha.commons.util.updater.ApplicationUpdater`)

The documentation in `org/opensha/commons/util/updater/README.md` has been updated accordingly.

kevinmilner (approved):
Looks good, I tested it on Linux and windows and especially like the new utility. Before merging, can you add documentation somewhere on how to run the utility? Maybe at the bottom of the Applications page of our website (edit it at https://github.com/opensha/opensha/wiki/Applications to show up at https://opensha.org/Applications)

abhatthal:
> Looks good, I tested it on Linux and windows and especially like the new utility. Before merging, can you add documentation somewhere on how to run the utility? Maybe at the bottom of the Applications page of our website (edit it at https://github.com/opensha/opensha/wiki/Applications to show up at https://opensha.org/Applications)

Sounds good. I'll add a section on the automatic update utility and then merge in this PR.

PR #241, opened 2026-08-03
NSHM23 model version support (adding revision 2)

kevinmilner:
This adds support for multiple versions of the NSHM23 model within a single ERF. This is needed because we are adding support for revision 2 of the model, and also want to support future model versions that include CEUS, Cascadia, and Alaska sources (and don't want to further pollute the ERF list with additional entries).

We don't want WUS-R2 to overwrite the current model (R1) via a getfile update; users require reproducibility of prior results, which is retained by adding R2 alongside R1. GetFile would only be used to update an existing model in place if a critical bug were discovered.

@abhatthal, please add the R2 file to the GetFile repos for NSHM23. It's file key is `WUS_branch_averaged_gridded_simplified_R2` and it can be downloaded from [here](https://data.opensha.org/nshm23/reports/results_WUS_FM_v3_branch_averaged_gridded_simplified_revised2026.zip). Once that has been set up, it should work with the code within this branch and we can update the default model to point to R2 instead of R1.

This PR also adds 2 new minor features used by the ERF:

* ERF Tectonic Regime change listener
  * new `org.opensha.sha.earthquake.BaseERF.addTectonicRegionChangeListener(ChangeListener)` method, allowing ERFs to notify the apps that their included TRTs have changed
  * This allows for the IMR GUI to update when a model's TRTs change, as will be the case when we include future NSHM23 models for subduction sources within this same NSHM23 ERF.
* `MultilineStringParameter`
  * New `StringParameter` that allows for displaying and editing multiline strings in a scrollable `JTextArea`
  * I use that here in order to display model information (non-editable)


abhatthal (approved):
LGTM.

- Updated the IMEventSetCalc CLT documentation to reflect the new ERF name.
- Updated the javadoc for constructors in `NSHM23_BranchAveragedERF`
- New ERF R2 model is deployed to both CARC and OpenSHA server GetFile backends
- Validated behavior with `HazardCurveApplication`. New MultilineStringParameter looks good, `IMRGuiBean` is responsive to change for NSHM23 ERF (listener working), and R2 data is retrieved from CARC lazily by `NSHM23Downloader` as expected.

kevinmilner:
> LGTM.
> 
>     * Updated the IMEventSetCalc CLT documentation to reflect the new ERF name.
> 
>     * Updated the javadoc for constructors in `NSHM23_BranchAveragedERF`
> 
>     * New ERF R2 model is deployed to both CARC and OpenSHA server GetFile backends
> 
>     * Validated behavior with `HazardCurveApplication`. New MultilineStringParameter looks good, `IMRGuiBean` is responsive to change for NSHM23 ERF (listener working), and R2 data is retrieved from CARC lazily by `NSHM23Downloader` as expected.

Good catch on the IM Event Set CLT README; I'll have to make corresponding updates in #243 as well. This brings up an interesting issue where the model is not select-able in the CLT; when we add more models, we may decide to try to add a special command line argument for this (or similar) ERF(s) to enable parameter adjustment. Not a show stopper now though, but something to revisit when we add CEUS and Alaska.

kevinmilner:
Updated the default solution to R2, will now merge.

PR #244, opened 2026-08-04
Increase application default sizing

kevinmilner:
The default sizes for the GUI applications are a holdover from days with much smaller and lower resolution displays. This PR increases the default application size to `1200 x 800` when it is able to detect the current maximum window size (primary screen size minus any space reserved for OS taskbars and such), and there are at least 50 leftover pixels available. If the available screen size is too small, it shrinks the window down accordingly, keeping the same aspect ratio, until is reaches the old default size.

Here is a screenshot of the Hazard Curve app with the old default sizing (`1000 x 720`):

image

...and with the new sizing (`1200 x 800`):

image

I apply it to the hazard curve, hazard spectrum, IM event set, attenuation relationship, and site data GUIs. I leave the GMT map and ShakeMap GUIs alone as they have smaller aspect ratios that already work fine.

@abhatthal, can you please check that everything looks good on a Mac?

This PR builds upon (and should not be merged prior to) #241.


abhatthal (approved):
LGTM.

Tested on macOS Sequoia 15.6.1 (aarch64), all impacted apps scale much better for a modern display.

I also tested on a UTM (qemu) Ubuntu VM where I scaled the display size / resolution. Impacted apps still look good on older displays.

Issue #207, opened 2026-03-02
GUI improvements for the next release


kevinmilner:
Reminder issue to work on GUI improvements for the next release, including:

* Pull #196 improved long IMR names, at least for Mac OS X, but doesn't seem to work on Linux (not sure about windows).
* Increase default size of applications; current sizes were chosen when monitors and resolutions were much smaller
* Investigate a larger look and feel refresh, possibly with [FlatLat](https://www.formdev.com/flatlaf/) which might be a pretty trivial fix to make our apps look more modern
* The bug report GUI doesn't look right dark mode, can't read any of the text

PR #259, opened 2026-08-14
Updated fix for IMR GUI long names on Linux

kevinmilner:
This attempts to finish the job started by #196 and mentioned in #207.

Long IMR names have historically been clipped in the selector dropdown box in our GUI apps. #196 fixed this on Mac's, but the issue remained on Linux and Windows.

This fixes it by:

* The application still chooses the collapsed IMR chooser's fixed width
* When the user clicks on the IMR name chooser, the popup begins opening triggering `PopupMenuListener.popupMenuWillBecomeVisible(PopupMenuEvent)`
* Detect that event and calculate a temporary size large enough for all IMR names.
* Temporarily set the chooser (the full GUI component) to the larger size by calling `setSize(...)`; the new size isn't acted upon (we don't trigger a layout update)
* Immediately schedule an event to revert to the original size via `SwingUtilities.invokeLater`.
* The popup then becomes visible and the window manager sees the temporarily-larger size of the combo box when determining the size for the popup. It sizes the popup relative to that larger size.
* Once the popup is visible, the event queue processes the revert command on the combo box, leaving the panel layout unchanged while the already-created popup remains wide enough.

Here's the app before/after clicking on the IMR chooser box:

image

And here is it with the chooser popup:

image

### DIFF
--- a/src/main/java/org/opensha/sha/gui/HazardCurveApplication.java
+++ b/src/main/java/org/opensha/sha/gui/HazardCurveApplication.java
@@ -1914,8 +1914,7 @@ ActionListener, ScalarIMRChangeListener, IMTChangeListener {
 
 		imrGuiBean = new IMR_MultiGuiBean(imrs);
 		imrGuiBean.addIMRChangeListener(this);
-		// imrGuiBean.setMaxChooserChars(30);
-        imrGuiBean.setChooserBoxSize(new Dimension(220, 25));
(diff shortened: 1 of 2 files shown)


kevinmilner:
Can confirm that the fix also works on Windows 11

image

abhatthal (approved):
LGTM. These fixes are already upstream in the Gradle loosened version restrictions PR I approved.

PR #260, opened 2026-08-15
Loosen gradle java version restrictions

kevinmilner:
We recently switched to using the modern gradle `toolchain` option which enforces a specific version of java. It's benefit is a more consistent environment, but it's downside is  that, if we target 21, you can't build even if you have a newer version (25) but not the exact version (21).

One way around this is to allow it to download its own java version via the `foojay` plugin. The annoyance there is that it needs to be enabled separately in each project (main, dev, cybershake, oaf, NZ), and is unnecessary if you already have a modern JDK.

Instead, this PR reverts to the old source/target compatibility flags because they're used by the eclipse plugin to determine the correct JDK, and also now specifies the java compiler `release` option. The latter is stricter than the old compatibility flags and should prevent accidental use of APIs beyond the current version.

More info on the various build flags is [available here](https://docs.gradle.org/current/userguide/building_java_projects.html).

### DIFF
--- a/build-common.gradle
+++ b/build-common.gradle
@@ -5,12 +5,25 @@
  * This file requires that the 'parentProject' property be defined (should be null for top level project)
  */
 
+
+/*
+ * Modern gradle recommends using toolchains but they come with an annoyance: you must have exactly the specified java
+ * version installed, or it fails. So if we specify Java 21 but you have 25, it will fail. That can be overcome by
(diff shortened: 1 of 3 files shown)


abhatthal (approved):
LGTM. IMR GuiBean renders nicely on macOS in the HazardCurveApp. Able to build with OpenJDK 24 locally after loosened version restrictions.

PR #266, opened 2026-09-06
JointRuptureExperimentalIMR: caching

voj:
In my previous PR, I wasn't too worried about performance. Now we have inversion solutions with several orders of magnitude more non-zero rates and hazard performance is becoming more important.

### DIFF
--- a/src/main/java/org/opensha/sha/earthquake/DistCachedERFWrapper.java
+++ b/src/main/java/org/opensha/sha/earthquake/DistCachedERFWrapper.java
@@ -57,7 +57,7 @@ public class DistCachedERFWrapper extends AbstractERF {
 	
 	private void initSources() {
 		List<ProbEqkSource> sources = new ArrayList<>();
-		Map<RuptureSurface, CustomCacheWrappedSurface> wrappedMap = new HashMap<>();
+		Map<RuptureSurface, RuptureSurface> wrappedMap = new HashMap<>();
 		for (ProbEqkSource source : erf) {
 			if (source instanceof FaultRuptureSource) {
 				RuptureSurface sourceSurf = getWrappedSurface(wrappedMap, source.getSourceSurface());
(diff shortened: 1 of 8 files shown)


src/main/java/org/opensha/sha/earthquake/faultSysSolution/util/SolHazardMapCalc.java
kevinmilner, line 1: I think it's worth keeping this option, we may run into a similar situation again down the road. Also, if the user sets the default surface caching policy to use `THREAD_LOCAL` caches, then the dist cache wrapper is no longer necessary.

So, I'd leave this boolean in there

src/main/java/org/opensha/sha/earthquake/DistCachedERFWrapper.java
kevinmilner, line 1: Good catch re: dropping the surface list.

kevinmilner (changes requested):
I'd leave the option to disable the cache-wrapped-ERF in `SolHazardMapCalc`; otherwise looks good.

And FYI, you might also try enabling the thread-local cache as an alternative to the distance-cache-wrapped version by calling this at the start of your main method:

`SurfaceCachingPolicy.force(CacheTypes.THREAD_LOCAL);`

It may have similar performance with less complication.

voj:
@kevinmilner 

I tried out `SurfaceCachingPolicy.force(CacheTypes.THREAD_LOCAL);` and it's a big improvement, but not quite as much as this PR. I often build a number of hazard maps at once, so it adds up. Here is an informal performance test for one map in minutes:

- **no caching**: 5:02
- **SurfaceCachingPolicy**: 2:01
- **this PR**: 1:20

src/main/java/org/opensha/sha/earthquake/faultSysSolution/util/SolHazardMapCalc.java
voj: done

kevinmilner (approved):

kevinmilner:
Good to know that it's still worth using the distance-cache wrapper!

